PR #21, opened 2026-03-06
Add MinGW cross-compilation support: make windows

rafael2k:
- Makefile: add 'windows' target for cross-building with MinGW from Linux
- os_interop.h: include <pthread.h> in _WIN32 block (MinGW winpthreads)
- telecom_system.cc: guard coarse_freq_sync block with MERCURY_GUI_ENABLED

PR #22, opened 2026-03-06
Fix headless build: decouple coarse_freq_sync from GUI state

xmutantson:
## Summary

- `telecom_system.cc` references `g_gui_state.coarse_freq_sync_enabled` outside of any `#ifdef MERCURY_GUI_ENABLED` guard, which breaks compilation without GLFW (as in PR #21's MinGW cross-build Makefile)
- Moves the flag to `cl_telecom_system` as a plain `bool` member so the physical layer has no dependency on GUI state
- Defaults to `false` (same as current behavior), set from INI at startup, bridged from GUI checkbox at runtime

## Changes

- **telecom_system.h**: Add `bool coarse_freq_sync_enabled` member
- **telecom_system.cc**: Initialize in constructor, use member instead of `g_gui_state`
- **main.cc**: Set from INI settings at startup, bridge from `g_gui_state` in GUI update loop

Related: #21

### DIFF
--- a/include/physical_layer/telecom_system.h
+++ b/include/physical_layer/telecom_system.h
@@ -193,6 +193,8 @@ public:
 
 	int bit_energy_dispersal_seed;
 
+	bool coarse_freq_sync_enabled;  // Coarse freq search (±30 Hz) for HF radio drift
+
 	st_reinit_subsystems reinit_subsystems;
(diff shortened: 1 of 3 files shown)


rafael2k:
Thanks!

PR #52, opened 2026-04-13
Fix TCP ARQ socket reads on Windows and add HAMLIB baud rate option

rafael2k:
WSAPoll() has known bugs on various Windows versions that can cause missed POLLIN events on connected sockets. This caused Mercury to accept ARQ client connections and send data, but never read incoming client data on Windows.

Replace WSAPoll with a select()-based hermes_poll() wrapper (Windows only). Add poll error checking and diagnostic logging around recv(). Fix broadcast recv_thread to use sock_errno() instead of perror().


pedromessetti:
LGTM

PR #54, opened 2026-04-15
* Patched MYCALL to allow up to 4 callsigns

n9seo:
MYCALL now supports up to 4 callsigns so it behaves more like vara.


n9seo:
@rafael2k Can you look at this when you have a moment? I tested this with K7EK last night and it should allow those of us who have several call signs listening on say BPQ to have the modem respond.  I just left a limit at 4 calls, 1 primary and 3 secondaries so that could be adjust but I feel 4 is a safe starting point.

n9seo:
I guess my branch got out of sync. I will rework this

PR #58, opened 2026-04-18
Added CALLINT command

n9seo:
This command allows us to adjust the timing between connection requests.

It is probably only useful to those of us using a BBS with mercury to adjust the wait time for a connection ACK before sending the next one.   If this window is too long (default) in some cases our radios may have changed frequency as we run systems that scan frequencies by changing the vfo every X seconds.  In my cause I use 7 for the scan duration as I found it works best with the now 5 modems here that are possible to use on my node/bbs.
So the pause CALLINT should be around 4-5 here in my case.

### DIFF
--- a/data_interfaces/tcp_interfaces.c
+++ b/data_interfaces/tcp_interfaces.c
@@ -344,6 +344,19 @@ static void execute_control_command(char *buffer)
         return;
     }
 
+    if (!memcmp(buffer, "CALLINT", strlen("CALLINT")))
+    {
+        memset(&cmd, 0, sizeof(cmd));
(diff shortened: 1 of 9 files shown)


rafael2k:
Thanks for this @n9seo ! This is pretty much appreciated. Let's review and merge asap.

rafael2k:
Thanks @n9seo. I'll merge this other PR, which I added some small fixes and unit tests https://github.com/Rhizomatica/mercury/pull/59

rafael2k:
Merged in https://github.com/Rhizomatica/mercury/commit/beec336f911765cf18caf22205c4184d2aaa2503

PR #76, opened 2026-06-06
broadcast: preserve payload length so VARA/KISS clients decode broadcasts

### LINKED ISSUE #75: Broadcast: received frames delivered to VARA/KISS clients keep the modem zero-padding (broadcasts never decode)
Nevermind, please disregard this issue post.

abogue77:
Fixes #75.

## Problem
When a VARA/KISS client (e.g. VarAC) uses the broadcast port (8100) as a TNC, **received broadcasts never decode**. The broadcast datalink is fixed-size, so a short AX.25 frame is zero-padded to `frame_size` on TX, and on receive Mercury forwards the whole padded frame (`frame_size - HEADER_SIZE` bytes) to the client, which rejects the trailing zeros. Transmit works; receive does not. (Full analysis in #75.)

Observed: VarAC sends a 40-byte broadcast (`KISS frame decoded: cmd=0x01 len=40`), the receiver's Mercury always delivers `Sending KISS frame to client: ... payload=125`, and the client drops it.

## Fix
Carry the real payload length in a 2-byte big-endian prefix after the 1-byte Mercury header, flagged with a header extension bit (`BCAST_EXT_LEN_PREFIX`):

- **TX** (`bcast_process_decoded_frame`): header + length prefix + payload, then zero-pad.
- **RX** (`bcast_get_tx_payload`): if the received frame's header carries the flag, return exactly the original payload with `CMD_AX25CALLSIGN`. This is detected from the **frame's own header**, not the latched `bcast_reply_cmd` (which defaults to `CMD_DATA` and resets on every client connect, so a receive-only station otherwise forwards the raw padded frame). Frames without the flag fall back to the existing header-strip.

`CMD_DATA` (hermes-broadcast) framing is unchanged.

## Tests
Updates the affected broadcast unit tests and adds a TX→RX round-trip test plus a receive-only regression test in `tests/data_interfaces/test_tcp_interfaces.c`. Full suite passes locally (`make test`).

## ⚠️ Compatibility note (needs a maintainer decision)
This **changes the on-air broadcast frame format**, so a patched station and a stock station will **not broadcast-interoperate** (ARQ is unaffected; a receiver still falls back to legacy for unflagged frames). A broadcast-format version bump or capability flag may be warranted — happy to adjust the approach (e.g. gate behind a version, or drive the AX.25-vs-CMD_DATA decision entirely off the frame header).

Verified end-to-end on two FT-857D/DigiRig stations: with both patched, broadcasts now decode into VarAC (`Async message received`).

### DIFF
--- a/data_interfaces/tcp_interfaces.c
+++ b/data_interfaces/tcp_interfaces.c
@@ -848,12 +848,28 @@ void *control_worker_thread_rx(void *conn)
 /********** BROADCAST TCP port INTERFACE **********/
 
 /*
+ * Length-prefixed broadcast framing.
+ *
+ * The broadcast datalink uses fixed-size modem frames, so a short VARA/AX.25
+ * frame is zero-padded up to frame_size for TX. Without recording the original
+ * length, the receiver can only hand the client the whole padded frame, and a
+ * VARA client (e.g. VarAC) then rejects it because the trailing zeros corrupt
+ * its payload. To fix this we store the real payload length in a 2-byte
+ * big-endian prefix right after the Mercury header and flag it with a header
+ * extension bit, so the receiver delivers the exact original frame. Receivers
+ * fall back to the legacy header-strip when the flag is absent.
+ */
+#define BCAST_LEN_SIZE       2     /* big-endian payload length after the header */
+#define BCAST_EXT_LEN_PREFIX 0x01  /* header extension bit: length prefix present */
+
+/*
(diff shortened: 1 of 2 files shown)


abogue77:
I want to do some more testing - abogue77

rafael2k:
> I want to do some more testing - abogue77

OK, no rush. Thanks!

PR #83, opened 2026-06-12
audioio: increase Linux ALSA/Pulse buffer from 30ms to 80ms

### LINKED ISSUE #81: Crackling noise with Pi/ALSA?
Hi guys,  I still don't know what to expect, but maybe you can tell me if this is normal.

I'm using "pat" winlink client in varahf mode and it's communicating with mercury api
just fine, but I'm getting maybe a 5% decode rate and cannot complete a transaction.

https://craiger.org/craiger/mercury-noise.mkv     << video/audio, with crackle, is this normal?

It sounds like a sample/bitrate issue I've seen before on some radios and rasperry
Pi, and it's present on the ic7300 and ic705.

I tried forcing 44.1Khz with .asoundrc, but that  produced pure noise.  is there
a way to have mercury use a sample rate other than 48000Hz?

Loaded configuration from mercury.ini
Audio System: Advanced Linux Sound Architecture (ALSA)
16:53:53.840 [+0.000s] [INF] [main] Async logger initialized (min_level=INFO)
16:53:53.840 [+0.000s] [INF] [main] Initializing I/O from Sound Card
16:53:53.841 [+0.001s] [INF] [radio-io] Initializing radio (type=3085, device=/dev/ttyACM0, hamlib_log_level=0, serial_speed=115200)
16:53:53.841 [+0.001s] [INF] [radio-io] Serial speed overridden to 115200 baud
16:53:53.920 [+0.080s] [INF] [audio-play] I/O playback (default) int32 / 48000Hz / 2ch / 42ms buffer
16:53:54.007 [+0.167s] [INF] [radio-io] Radio control: HAMLIB (model 3085, device /dev/ttyACM0)
16:53:54.007 [+0.167s] [INF] [main] Initializing Modem
16:53:54.007 [+0.167s] [INF] [modem] Created data buffers for ARQ and BROADCAST datalink, tx/rx paths.
16:53:54.009 [+0.169s] [INF] [modem] Initialized persistent FreeDV mode pool (DATAC13/DATAC4/DATAC3/DATAC1), frames per burst: 1
TX test: bytes_per_modem_frame=128, payload_size=126
16:53:54.009 [+0.169s] [INF] [modem] Active FreeDV mode at startup: 12 (DATAC3), verbosity: 0
16:53:54.009 [+0.169s] [INF] [modem] Modem expects sample rate: 8000 Hz
16:53:54.009 [+0.169s] [INF] [modem] Modem payload bytes per frame: 126
16:53:54.009 [+0.169s] [INF] [modem] Split control/data mode switching: ENABLED
16:53:54.022 [+0.182s] [INF] [radio] TX enabled (PTT ON)
16:53:54.048 [+0.208s] [INF] [audio-cap] I/O capture (default) int32 / 48000Hz / 2ch / 42ms buffer
16:53:57.895 [+4.055s] [INF] [radio] TX disabled (PTT OFF)
16:53:57.909 [+4.069s] [INF] [radio] TX enabled (PTT ON)
^X16:54:01.782 [+7.942s] [INF] [radio] TX disabled (PTT OFF)
16:54:01.797 [+7.957s] [INF] [radio] TX enabled (PTT ON)
^C16:54:05.241 [+11.401s] [INF] [main] Termination signal received, shutting down...
16:54:05.248 [+11.408s] [ERR] [audio-cap] ffaudio.stop: snd_pcm_pause: (-5) Input/output error
16:54:05.312 [+11.472s] [INF] [audio-cap] radio_capture_thread exit
16:54:05.440 [+11.600s] [INF] [audio-play] radio_playback_thread exit

rafael2k:
Raspberry Pi USB audio suffers from isochronous transfer jitter that causes buffer underruns with small 10ms periods. The resulting gaps in TX audio manifest as crackling noise that prevents reliable decoding by remote stations (e.g. VARA HF at ~5% decode rate).

Both ALSA and PulseAudio backends share this buffer length config. Increasing to 80ms (3 periods of ~27ms) gives the Pi's dwc_otg USB driver sufficient headroom to avoid underruns.

Closes #81

### DIFF
--- a/audioio/audioio.c
+++ b/audioio/audioio.c
@@ -413,7 +413,7 @@ void *radio_playback_thread(void *device_ptr)
     if (audio_subsystem == AUDIO_SUBSYSTEM_DSOUND)
         audio = (ffaudio_interface *) &ffdsound;
 #elif defined(__linux__)
-    conf.buf.buffer_length_msec = 30;
+    conf.buf.buffer_length_msec = 80;
     period_ms = conf.buf.buffer_length_msec / 3;
     if (audio_subsystem == AUDIO_SUBSYSTEM_ALSA)
         audio = (ffaudio_interface *) &ffalsa;
@@ -689,7 +689,7 @@ void *radio_capture_thread(void *device_ptr)
         audio = (ffaudio_interface *) &ffdsound;
     }
 #elif defined(__linux__)
-    conf.buf.buffer_length_msec = 30;
+    conf.buf.buffer_length_msec = 80;
     if (audio_subsystem == AUDIO_SUBSYSTEM_ALSA)
         audio = (ffaudio_interface *) &ffalsa;
     if (audio_subsystem == AUDIO_SUBSYSTEM_PULSE)


rafael2k:
Still open from July. CI is green and the change itself is two lines, but I do not think green CI is the relevant evidence here, so I am not merging it on that basis.

Raising the Linux ALSA/Pulse buffer 30 ms → 80 ms adds up to 50 ms of latency in **each** direction, and on the playback path it also moves `period_ms` from 10 ms to ~26 ms. That latency lands directly in the ARQ turnaround budget, which is measured in hundreds of milliseconds and which this project has already been bitten through twice from the audio/PTT side (the `post_ptt_delay` field bug, and the Windows waterfall loop holding PTT an extra 1.1 s so every ACK collided with our own tail).

We also have a standing rule not to tune ARQ guard intervals against unmeasured audio-backend buffering, and this is the same quantity approached from the other end.

What would settle it is one measurement, not a review: the same transfer on the same link with 30 ms and with 80 ms, reporting turnaround and total time. If the crackle goes away and the turnaround does not move, it merges as-is. If the turnaround moves, we know the cost we are paying for the fix and can decide with the number in hand.

Holding until then.

rafael2k:
Still holding, and I want to be explicit that this is not neglect — I tried to settle it today and established that it **cannot** be settled from the bench.

My earlier ask was one measurement: the same transfer with 30 ms and with 80 ms, reporting turnaround. I went to run that on the loopsim harness and the harness cannot produce it. The two lines this PR changes sit under `AUDIO_SUBSYSTEM_ALSA` / `AUDIO_SUBSYSTEM_PULSE`, and the integration harness runs the FIFO/file audio backends — so on the bench these lines are simply not executed. Any number I produced there would be noise dressed up as evidence.

That leaves the field stations as the only instrument, on both halves of the question:

- **Does it fix the crackle?** Only reproducible on Pi USB audio with real isochronous jitter.
- **What does it cost?** Up to 50 ms each way, and `period_ms` 10 ms → ~27 ms on playback, landing directly in an ARQ turnaround budget measured in hundreds of ms.

One `uucp` transfer at 30 ms and one at 80 ms between estacao2 and estacao3, with the `[arq-timing]` records from each, answers both at once.

Two things have also changed under this PR since July that will matter when it does get tested:

- It is now **427 commits behind**. It still applies cleanly, but `audioio.c` has been reworked around it — the file now has several `buffer_length_msec` sites with different values (40 as the general default, 200 elsewhere), so the two 30 ms Linux constants are no longer the obviously-uniform thing they were when this was written. Worth re-reading the surrounding code before measuring, in case the right fix is now shaped differently.
- CI has grown TSan, ASan+UBSan and a macOS build since these checks ran; the green ticks above predate all three.

The change may well be correct — the underrun reasoning is plausible and #81 is a real complaint. I just will not merge a latency change into the ARQ path on plausibility, having been bitten twice from exactly this direction (`post_ptt_delay` silently eating turnaround, and the Windows waterfall loop holding PTT an extra 1.1 s so every ACK collided with our own tail).

Rebasing it and running the pair of transfers unblocks it immediately.

PR #85, opened 2026-06-17
audioio: right-size resampler scratch buffers (fixes #79)

### LINKED ISSUE #79: Segmentation fault  (pat/winlink)
Hi all - just getting started.  mercury compiled mercury-qt connected - great!

Just trying with Pat client on default port 8300 configured as "vara hf" - should this work?

strace....
```
14:39:34.224 [+12.845s] [INF] [tcp-ctl] Control client connected
14:39:34.225 [+12.847s] [INF] [tcp-data] Data client connected
14:39:34.226 [+12.847s] [INF] [tcp-ctl] Command received: PUBLIC ON
14:39:34.226 [+12.847s] [INF] [tcp-ctl] Command received: CWID ON
14:39:34.226 [+12.847s] [WRN] [tcp-ctl] Unknown command: CWID ON
14:39:34.226 [+12.848s] [INF] [tcp-ctl] Command received: COMPRESSION TEXT
14:39:34.226 [+12.848s] [INF] [tcp-ctl] Command received: MYCALL KM6LYW
14:39:34.226 [+12.848s] [INF] [tcp-ctl] Command received: LISTEN OFF
14:39:34.226 [+12.848s] [INF] [tcp-ctl] Command received: VERSION
14:39:34.228 [+12.849s] [INF] [arq] My callsign: KM6LYW
14:39:34.228 [+12.849s] [INF] [tcp-ctl] Command received: BW2300
14:39:34.228 [+12.849s] [INF] [tcp-ctl] Command received: BW2300
14:39:34.228 [+12.849s] [INF] [tcp-ctl] Command received: WINLINK SESSION
14:39:34.228 [+12.849s] [WRN] [tcp-ctl] Unknown command: WINLINK SESSION
14:39:34.228 [+12.849s] [INF] [tcp-ctl] Command received: CONNECT KM6LYW KG6MDW
14:39:34.423 [+13.045s] [INF] [radio] TX enabled (PTT ON)
+++ killed by SIGSEGV +++
**Segmentation fault**
```

Raspberry Pi armhf,   DigiPi  project digipi.org

[main]
ui_enabled = true
ui_port = 10000
ui_protocol = ws
waterfall_enabled = true
radio_model = 3073
radio_device = "/dev/ttyUSB0"
input_device = "plughw:0,0"
output_device = "plughw:0,0"
capture_channel = left
sound_system = alsa
arq_tcp_base_port = 8300
broadcast_tcp_port = 8100
verbose = false
freedv_verbosity = 0
hamlib_log_level = 0
radio_serial_speed = 19200

[arq]
no_progress_timeout_s = 180
disconnect_drain_timeout_s = 30

[audio]
tx_gain_db = -1.00

thagale:
# audioio: right-size resampler scratch buffers (fixes #79)

## Summary

`radio_playback_thread` and `radio_capture_thread` size their resampler scratch
buffers from `SIGNAL_BUFFER_SIZE` (12,288,000 samples) — roughly **590 MB for the
playback stereo buffer alone and ~1 GB across both threads**. On low-RAM 32-bit
hosts — notably the **512 MB Raspberry Pi Zero2W** — these allocations fail, so
the modem cannot run there. Hosts with more RAM (e.g. a Pi4) allocate
successfully and never hit it, matching @craigerl's observation that the Pi4 is
fine while the Zero2W crashes.

`728dbf7` (the recent correctness batch) added `NULL` checks on exactly these
allocations, so this PR drops the duplicate checks and is now **resize-only**,
stacking on top of that change. Note the `NULL` check turns the failure from a
segfault into a graceful `Failed to allocate ... buffers` exit — but the
allocation still fails, the audio thread still can't start, and the board stays
unusable. **Right-sizing is what actually lets it run.**

These buffers only ever hold **one audio period**, so this PR sizes them from
the period (playback: `period_ms`; capture: the device buffer length). The
per-thread footprint drops from ~1 GB to a few KB.

## Root cause detail

Before `728dbf7` the unchecked 590 MB `malloc` returned `NULL` and was
dereferenced on the first TX, producing the #79 segfault. This is confirmed by
@craigerl's attached core dump: the crashing thread is in
`radio_playback_thread`, faulting on the mono→stereo write (`str r12, [r1]` with
`r1 = 0x0`), i.e. `buffer_output_stereo[idx] = ...` with `buffer_output_stereo ==
NULL`. `buffer_upsampled` (295 MB) is a valid pointer in the same core, so only
the largest (590 MB) allocation failed — exactly the `{×1, ×6, ×12}` oversizing.
That also shows the 512 MB board returns `NULL` here (not an OOM-kill).

The loop reads at most one 8 kHz period and upsamples 1:6 to stereo, so the live
data is a few hundred samples — the `SIGNAL_BUFFER_SIZE` sizing over-allocated by
~10^4–10^5×.

## The fix

- **Playback:** `period_scratch_bytes = 8000 * sizeof(int32_t) * period_ms / 1000
  + 64`, then size `input_buffer` / `buffer_upsampled` (×6) / `buffer_output_stereo`
  (×12) from it.
- **Capture:** size `buffer_output` / `buffer_downsampled` from
  `cfg->sample_rate * cfg->buffer_length_msec / 1000` (+50% margin).
- The `728dbf7` `NULL` checks are kept.

No functional change to the audio path — the resampler is stateful across
periods and these buffers are per-iteration scratch; correct sizing produces
identical output.

## Verification

- Builds clean with `gcc` on armhf (Debian Trixie) against current `mercuryv2`.
- **Reproduction (emulated armhf, 32-bit):** an `LD_PRELOAD` shim returning `NULL`
  for allocations ≥ 400 MB (simulating the Zero2W's refusal of the 590 MB
  request) makes the pre-`728dbf7` build **segfault at PTT-ON**; current
  `mercuryv2` fails gracefully but the audio thread exits; **with this patch the
  modem runs** (the giant allocation is gone, nothing to refuse).
- **AddressSanitizer** over the real `-x alsa` playback path (CONNECT/TX) reports
  no overflow against the resized buffers — the new sizing is correct.

Confirmation on a physical 512 MB Pi Zero2W by @craigerl would be the ideal final
check.

### DIFF
--- a/audioio/audioio.c
+++ b/audioio/audioio.c
@@ -452,14 +452,22 @@ void *radio_playback_thread(void *device_ptr)
     ffuint frame_size;
     ffuint msec_bytes;
 
-    // input is int32_t (8kHz samples from playback_buffer)
-    int32_t *input_buffer = (int32_t *) malloc(SIGNAL_BUFFER_SIZE * sizeof(int32_t));
+    /* Per-period resampler scratch.  These only ever hold ONE 8 kHz read
+     * period (period_bytes_8k below) and its 1:6 upsampled 48 kHz stereo
+     * expansion.  Sizing from SIGNAL_BUFFER_SIZE allocated ~590 MB for the
+     * stereo buffer alone (~1 GB across both audio threads), which fails on
+     * low-RAM 32-bit hosts such as the 512 MB Pi Zero2W — the modem could not
+     * run there (issue #79).  Size from the actual period instead. */
+    size_t period_scratch_bytes = (size_t) 8000u * sizeof(int32_t) * period_ms / 1000u + 64;
 
-    // upsampled buffer (48kHz mono)
-    int32_t *buffer_upsampled = (int32_t *) malloc(SIGNAL_BUFFER_SIZE * sizeof(int32_t) * 6);
+    // input is int32_t (8kHz samples from playback_buffer)
+    int32_t *input_buffer = (int32_t *) malloc(period_scratch_bytes);
+
+    // upsampled buffer (48kHz mono), 1:6 upsample
+    int32_t *buffer_upsampled = (int32_t *) malloc(period_scratch_bytes * 6);
 
     // output is int32_t stereo (48kHz)
-    int32_t *buffer_output_stereo = (int32_t *) malloc(SIGNAL_BUFFER_SIZE * sizeof(int32_t) * 2 * 6); // a big enough buffer
+    int32_t *buffer_output_stereo = (int32_t *) malloc(period_scratch_bytes * 2 * 6);
 
     if (!input_buffer || !buffer_upsampled || !buffer_output_stereo)
     {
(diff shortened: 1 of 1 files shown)


thagale:
Rebased onto current `mercuryv2` after the correctness batch landed — thanks for `728dbf7`, which added `NULL` checks on these same allocations, so I've dropped mine. **The PR is now resize-only** and stacks on top of that change.

One thing worth flagging: the `NULL` check turns the Zero2W failure from a segfault into a graceful `Failed to allocate ... buffers` exit, but the allocation itself is still ~590 MB for the playback stereo buffer (~1 GB across both audio threads). So on a 512 MB board it still fails and the audio thread can't start — the modem stays unusable there. Since these buffers only ever hold one audio period, this sizes them from the period (playback `period_ms`, capture device-buffer length), dropping the footprint to a few KB while keeping your `NULL` checks.

rafael2k:
Thanks for fix @thagale !

craigerl:
Thank you guys!  I still need to test on Raspberry Pi Zero2W  (512MB)... been busy

PR #89, opened 2026-06-23
build: fix macOS (Apple clang) compilation

thagale:
Mercury doesn't build on macOS (Apple clang) today. This fixes the three glibc-only spots that stop it: the `exp10f` use in FreeDV's `POW10F` macro, and `<malloc.h>` / `memalign` / `SHMLBA` in `ring_buffer_posix.c`. All changes are behind `__APPLE__`, so Linux and Windows builds are unaffected. Verified building on both macOS and Linux/armhf, with a working binary on each.

Scope note: this fixes the build. It doesn't claim validated CoreAudio runtime on Mac; the control/TNC paths run, but I didn't test audio on Mac hardware.

Diff is 2 files, +16/-1.

### DIFF
--- a/common/ring_buffer_posix.c
+++ b/common/ring_buffer_posix.c
@@ -20,7 +20,22 @@
 #include <string.h>
 
 #include <time.h>
+#if defined(__APPLE__)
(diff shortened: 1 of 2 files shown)


rafael2k:
Thanks!

PR #93, opened 2026-06-30
Prevent integer overflow in playback duration calculation

### LINKED ISSUE #90: Possible integer overflow in modem.c
While debugging on NixOS, I found that the TX process gets stuck.
After tracing the execution (with some assistance from LLM), I noticed that the following line in modem.c appears to overflow:

```
/* Wait for all samples to be played out */
usleep(total_samples * 1000000 / FREEDV_FS_8000);
```

Image

Interestingly, I only observed this issue on NixOS. It appears to work correctly on my friend's Debian machine, so this may be due to differences in compiler versions, optimization settings, or even hardware.

Since signed overflow in C is undefined behavior, would it make sense to explicitly promote the calculation to 64-bit arithmetic? For example:
`usleep(total_samples * 1000000ULL / FREEDV_FS_8000);`

I'm not very familiar with C or NixOS internals, so I may be missing something. If this is expected behavior, please let me know.

Looking forward to your reply.

pedromessetti:
Fix undefined behavior bug where multiplying large burst sizes as standard integers caused an arithmetic overflow before division.

Fix #90

### DIFF
--- a/modem/modem.c
+++ b/modem/modem.c
@@ -713,7 +713,7 @@ int send_modulated_data(generic_modem_t *g_modem, uint8_t *bytes_in, int frames_
         return -1;
     }
 
-    int total_samples = 0;
+    size_t total_samples = 0;
     float tx_gain = atomic_load(&g_tx_gain);
     float peak_fs = 0.0f;  /* pre-saturation peak magnitude across this burst */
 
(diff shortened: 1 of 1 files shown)


rafael2k (approved):
LGTM

PR #94, opened 2026-07-03
arq: fix permanent deadlock when app TX ring fills during transfer

thagale:
## What breaks and why

The ARQ event loop wedges permanently after about 64 KB of data has been pushed
into a session. On a Winlink email transfer this happens within seconds, since
the HF link drains at a few hundred bytes per second but the TCP client writes at
memory speed.

The root cause is a lock-ordering bug in `arq_payload_bridge_worker`
(`datalink_arq/arq.c`). The bridge worker thread holds `g_app_tx_mtx` across a
call to `write_buffer()`. When the 64 KB `g_app_tx_buf` ring fills,
`write_buffer()` blocks in `COND_WAIT` — but COND_WAIT releases only the ring's
own internal mutex, not the outer `g_app_tx_mtx`. The only code that drains the
ring is `cb_tx_read` (the FSM's `.tx_read` callback, called from the ARQ event
loop thread), which must first acquire `g_app_tx_mtx`. So the ring never drains,
the COND_WAIT never fires, and the event loop is permanently wedged.

The same pattern was in `arq_queue_data` (the alternative direct write path).

## The fix

Two changes, tightly coupled:

**`datalink_arq/arq.c` (+0/-4):** Remove `g_app_tx_mtx` from both write sites.
The ring already has its own `pthread_mutex_t` + `pthread_cond_t` (initialized
in `circular_buf_init`, line 773). Concurrent read and write are fully serialized
by the ring's internal lock; the outer lock was redundant and caused the deadlock.

**`common/ring_buffer_posix.c` (+1/-0):** Add `COND_SIGNAL` to `clear_buffer()`.
Without the signal, any writer blocked in `write_buffer()`'s COND_WAIT stays
blocked even after the ring is cleared on session teardown, causing a shutdown
hang as a secondary symptom.

## What was not changed

All read-side uses of `g_app_tx_mtx` (`cb_tx_read`, `cb_tx_backlog`) and the
`clear_buffer` call sites in `arq.c` are unchanged. The outer mutex at the read
side causes no deadlock and its presence or absence there does not affect this fix.

## Verification

- Builds clean on macOS (Apple clang) and Linux armhf (gcc).
- New unit test in `tests/common/test_ring_buffer.c`: fills a ring to capacity,
  parks a writer thread in `write_buffer()`, calls `clear_buffer()` from the main
  thread, and joins the writer. The test hangs permanently if `clear_buffer()`
  does not signal — making a regression immediately visible. Passes with the fix.
- `tests/Makefile`: drop `-lrt` on macOS (not present there; `clock_gettime` etc.
  are in libc on macOS 10.12+), so the test suite builds on both platforms.

Diff: 4 files, +66/-4 lines.

### DIFF
--- a/common/ring_buffer_posix.c
+++ b/common/ring_buffer_posix.c
@@ -355,6 +355,7 @@ void clear_buffer(cbuf_handle_t cbuf)
     cbuf->internal->head = 0;
     cbuf->internal->tail = 0;
 
+    COND_SIGNAL( &cbuf->internal->cond );
     MUTEX_UNLOCK( &cbuf->internal->mutex );
 }
 
--- a/datalink_arq/arq.c
+++ b/datalink_arq/arq.c
@@ -497,9 +497,7 @@ static void *arq_payload_bridge_worker(void *arg)
     {
         if (payload.len == 0 || payload.len > INT_BUFFER_SIZE)
             continue;
-        pthread_mutex_lock(&g_app_tx_mtx);
         write_buffer(g_app_tx_buf, payload.data, payload.len);
-        pthread_mutex_unlock(&g_app_tx_mtx);
 
         arq_event_t ev = { .id = ARQ_EV_APP_DATA_READY };
         evq_push(&ev);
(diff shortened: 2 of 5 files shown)


rafael2k:
Thanks @thagale ! I'll take a look ASAP.

PR #109, opened 2026-07-07
arq: make ARQ timing constants INI-tunable

thagale:
Eight ARQ timing constants are currently hardcoded `#define`s. Operators running Mercury over unusual paths (long-distance HF, aircraft, marginal propagation) cannot tune them without recompiling. This makes them `_Atomic int` globals with clamped INI keys under `[arq]`.

**Constants added:**

| INI key | Default | Range | Notes |
|---|---|---|---|
| `channel_guard_ms` | 700 | 200–3000 | Inter-frame gap |
| `iss_post_ack_guard_ms` | 900 | 200–3000 | Post-ACK quiet time |
| `keepalive_interval_s` | 20 | 5–120 | |
| `keepalive_miss_limit` | 5 | 2–20 | Misses before disconnect |
| `ladder_up_successes` | 2 | 1–16 | Consecutive ACKs to step up |
| `retry_downgrade_threshold` | 2 | 1–16 | Retries to step down |
| `mode_hold_after_downgrade_s` | 6 | 0–60 | Hold period after step-down |
| `data_retry_slots` | (existing) | 1–32 | Was already atomic; INI key added |

All defaults match the current hardcoded values — no behavior change at defaults. `no_progress_timeout_s` and `disconnect_drain_timeout_s` are already atomic upstream and are untouched.

Each constant follows the pattern of the existing `_Atomic` globals in `arq_protocol.h`. Setter functions clamp on write; `cfg_utils.c` validates on parse. `mercury.ini.example` documents all eight keys.

Adds `tests/common/test_cfg_utils.c` (defaults, INI round-trip, out-of-range clamp rejection).

### DIFF
--- a/common/cfg_utils.c
+++ b/common/cfg_utils.c
@@ -54,6 +54,14 @@ void cfg_set_defaults(mercury_config *cfg)
     cfg->tnc_keepalive_s = 60;
     cfg->tnc_buffer_report_ms = 1000;
     cfg->tx_gain_db            = 0.0f;
+    cfg->data_retry_slots            = ARQ_DATA_RETRY_SLOTS_DEFAULT;
+    cfg->mode_hold_after_downgrade_s = ARQ_MODE_HOLD_AFTER_DOWNGRADE_S_DEFAULT;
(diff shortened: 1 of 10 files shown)


rafael2k:
Applied.

PR #110, opened 2026-07-07
arq: replace extern-in-.c TNC coupling with arq_tnc_callbacks_t

thagale:
`arq.c` currently declares six TNC notification functions via a bare `extern` block inside the .c file. This creates an upward dependency — the ARQ layer reaches into the TNC layer at link time with no header, no interface contract, and no way to substitute an alternative implementation for testing.

This PR introduces `datalink_arq/arq_tnc.{h,c}`, a small registered-callback seam matching the pattern already used for the FSM callbacks (`arq_fsm_callbacks_t`).

**What changes:**
- New `arq_tnc_callbacks_t` struct with 6 function pointers matching the exact signatures of the `tnc_send_*` functions
- `arq_set_tnc_callbacks()` registration function; invokers are NULL-guarded
- `arq.c`: extern block removed; all 10 call sites converted to dispatch through the seam
- `tcp_interfaces.c`: populates a static callback table and registers it as the first line of `interfaces_init()`
- `init_model()` is left as a plain extern (out of scope here)

No behavior change at runtime. All 10 call sites converted. Clean build, 28 tests pass including 4 new `test_arq_tnc` unit tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ LDFLAGS=$(FFAUDIO_LINKFLAGS) -lm $(HAMLIB_LDFLAGS)
 
 MERCURY_LINK_INPUTS = \
 	main.o common/cfg_utils.o common/iniparser/iniparser.o common/iniparser/dictionary.o \
-	datalink_arq/arq.o datalink_arq/arith.o datalink_arq/arq_channels.o \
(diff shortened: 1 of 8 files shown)


rafael2k:
Applied. Thanks @thagale !

PR #112, opened 2026-07-07
arq: fix S1 fade-cliff regression — mode downgrade blocked by stuck retry counter and dead-code gate

thagale:
When the channel fades completely, Mercury should step down the mode ladder after `ARQ_HARD_LOSS_THRESHOLD` (8) consecutive failed bursts. Instead it retries indefinitely at the current mode. Two bugs compound to cause this:

**Bug 1 — retry counter stuck at 2**

In the WAIT_ACK retry-exhaustion path, `consecutive_retries` was clamped to `ARQ_RETRY_DOWNGRADE_THRESHOLD` (2) but never incremented past it. Without any ACK, the counter never reached `ARQ_HARD_LOSS_THRESHOLD` (8), so the hard-loss downgrade never fired.

Fix: increment `sess->consecutive_retries` before the clamp so it accumulates across budget cycles.

**Bug 2 — window gate made hard-loss bypass unreachable**

`maybe_upgrade_mode` returned false unconditionally when `tx_window_count > 0`. On a dead channel there is always a frame in the window, so the hard-loss bypass code below the gate was dead code regardless of how high `consecutive_retries` grew.

Fix: tighten the gate to `tx_window_count > 0 && consecutive_retries < ARQ_HARD_LOSS_THRESHOLD`, so the hard-loss path bypasses the window check once the threshold is reached.

**Test**

`tests/sim/test_arq_sim.c::test_sim_fade_cliff_downgrades` was previously `TEST_IGNORE`d (the harness existed but the test body was a placeholder). It is now a full regression test: drives ISS timers manually with zero ACK delivery through 7 budget exhaustions, intercepts the resulting `MODE_REQ` for `DATAC15`, injects a synthetic `MODE_ACK`, and asserts `payload_mode == DATAC15`. All 135 tests pass.

### DIFF
--- a/datalink_arq/arq_fsm.c
+++ b/datalink_arq/arq_fsm.c
@@ -503,8 +503,15 @@ static bool maybe_upgrade_mode(arq_session_t *sess)
      * wrong byte count (a FULL DATAC15 frame read at DATAC3's larger slot
      * over-delivers: the 118-vs-112 bidirectional regression).  Defer the change
      * until the window drains; it reaches 0 after every fully-ACKed burst, so
-     * mode adaptation still happens at burst boundaries. */
-    if (sess->tx_window_count > 0)
+     * mode adaptation still happens at burst boundaries.
+     *
+     * Exception: hard-loss floor drop (consecutive_retries >= ARQ_HARD_LOSS_THRESHOLD).
+     * On a total-fade channel the window never drains because no ACK ever arrives,
+     * so deferring until burst-boundary means the downgrade never happens.  Allow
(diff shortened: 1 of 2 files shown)


thagale:
After reviewing PR #103 more carefully, Rafael's fix here is more complete — it addresses the root cause (bounding the reverse-hold streak so the peer_snr_supports_mode guard doesn't silently prevent retry accumulation), adds a restage mechanism to avoid data loss on mode changes, and includes a much more thorough fading fuzz suite. Our fix here addressed two downstream symptoms of the same root cause. Closing this in favor of #103.

PR #120, opened 2026-07-10
Fyne UI with CGo + Windows Installer

pedromessetti:


rafael2k:
## Review — Fyne UI with CGo + Windows Installer

Solid, pragmatic feature. It embeds the Mercury C engine in-process via a CGo bridge, plus a Windows Inno Setup installer, and **changes no existing C modem/ARQ source** — only `Makefile`, `release.yml`, and new files under `gui_interface/fyne-ui/` + `windows-installer/`. So it's **safe for the running stations** (zero risk to the modem/ARQ path). Nice touches: clean `mercury_embedded` build-tag split with a stub fallback, Go unit tests for the wire parsers, and the teardown correctly does `shutdown_modem` **before** `audioio_deinit` (respects the known shutdown-UAF ordering).

Findings, by priority:

### Should-fix before/around merge (correctness)

**(1) Teardown leaks the TCP threads — `interfaces_shutdown()` is missing.**
`main.c` teardown calls `interfaces_shutdown()` (main.c:712), *before* `shutdown_modem`. `mercury_shutdown()` in `mercury_bridge.c` never calls it, so the control/data/broadcast server threads keep running and the ports stay bound after shutdown → a second `mercury_init()` (UI restart / device change) would fail or hit a use-after-free. Add `interfaces_shutdown()` as the first teardown step.

**(2) It re-introduces the retry-slot regression fixed in #119.**
`mercury_bridge.c:144` copies `arq_set_retry_slots(g_mcfg.data_retry_slots)` — the exact line that inflates CALL/ACCEPT retries and caused the ~90 s `ACCEPTING` linger in the recent OTA. Once #119 lands this must become `arq_set_data_retry_slots(...)`. (Direct consequence of item 6.)

### CGo/runtime hazards (should-fix)

**(3) C-side `signal(SIGINT/SIGTERM)` + `alarm(10)` inside a Go process.**
`main.c` can own signals because it *is* the process; a CGo library must not — the Go runtime installs and manages its own signal handlers, and overriding them (plus a bare `alarm(10)` with no `SIGALRM` handler, which will hard-kill the whole Fyne app) is a classic CGo footgun. Prefer: let Go own signals (`os/signal.Notify`) and call `mercury_request_shutdown()`; drop the C handlers and the `alarm()` watchdog (or move the watchdog to Go).

### Build correctness (minor)

**(4) `libmercury_core_w64.a` target archives into `libmercury_core.a`** (name mismatch) → it never satisfies its own target so it rebuilds every time, and it clobbers the native `libmercury_core.a`. Combined with extracting freedv/audioio `.o` into the shared `modem/freedv` & `audioio` dirs, native and mingw objects can cross-contaminate (the known mingw-`.o` gotcha) — a `clean` is required between native and Windows builds. Give the w64 archive its own filename and its own extraction dir.

**(5) No native (Linux) `fyne-ui` make target**, and the UI isn't in `all` or CI — the Linux build path is manual/untested. `windows-zip`/release now also hard-depend on the Go+Fyne+mingw toolchain (worth noting for the release runner).

### Architecture (the "can be improved")

**(6) `mercury_init()` is a ~120-line copy-paste of `main.c`'s init** → guaranteed drift; #119 and the busy-detector wiring already have to be mirrored by hand. Extract a shared `mercury_engine_init(const mercury_config*)` (and matching teardown) that **both** `main.c` and the bridge call. This single change also prevents items 1 and 2 from ever recurring.

**(7) In-process engine, but the UI talks to it over a loopback WebSocket** (`buildWebSocketURL`/`parseSpectrumFrame` connect to the engine's own WS server on 127.0.0.1). It works and reuses the existing UI protocol, so fine for v1 — a direct CGo data path would drop the localhost hop later.

**Bottom line:** mergeable as an isolated, station-safe feature, but items 1 (thread leak) and 2 (retry regression) should be fixed first, with item 3 (Go signal ownership) close behind. Item 6 is the real long-term improvement.

rafael2k:
## Re-review after the latest 3 commits (`c53684a`)

Thanks — most of the review is addressed. Status:

- ✅ **(1) `interfaces_shutdown()`** now called as the first teardown step (correct order, before `shutdown_modem`).
- ✅ **(3) Signals** — C `signal()` handlers and `alarm(10)` removed; shutdown moved to Go (`os/signal.Notify` → `window.Close()`). 👍
- ✅ **(4, contamination half)** — the `ar x` extract-and-repack is gone; freedv/audio now link directly via the CGo `LDFLAGS` (`-lfreedvdata -l:audioio.a`). Cleaner, and it removes the native/mingw `.o` cross-contamination risk.
- ✅ **(5)** — native Linux `fyne-ui:` target added.

### Remaining

**(A) The retry-slot regression is back — blocked on a rebase.**
`37cc114` correctly switched to `arq_set_data_retry_slots(...)`, but `90203fe` reverted it to `arq_set_retry_slots(...)`. Cause: this branch's base is `8b7600b`, whereas `mercuryv2` has since advanced and now contains #119 (which introduces `arq_set_data_retry_slots`). So the correct function doesn't exist on this branch yet.
→ **Rebase `cgo` onto current `mercuryv2`, then set `mercury_bridge.c` back to `arq_set_data_retry_slots(g_mcfg.data_retry_slots)`.** Otherwise the integrated UI re-inflates CALL/ACCEPT retries (the ~90 s `ACCEPTING` linger).

**(B) `libmercury_core_w64.a` target still archives into `libmercury_core.a`** (name mismatch). It never satisfies its own target (so it rebuilds every time) and it clobbers the native `libmercury_core.a`, so a native + Windows build without a `clean` in between mixes architectures.
→ Emit `libmercury_core_w64.a` (its own filename) and point `fyne-ui-windows` at it. Minor.

**(C) Architecture (optional, longer-term):** `mercury_init()` is still a ~120-line copy of `main.c`'s init, so fixes like #119 and the busy-detector wiring have to be mirrored by hand (exactly what happened with item A). Extracting a shared `mercury_engine_init(const mercury_config*)` used by both `main.c` and the bridge would prevent this class of drift. Not a blocker for this PR.

Nothing here touches the modem/ARQ core, so it stays station-safe. Items A and B are quick; A should land before merge.

rafael2k:
## Re-review after the rebase + engine refactor (`0f36025`) — and a fix pushed (`c3cd9ee`)

Nice work on the refactor — extracting `common/mercury_engine.{c,h}` so `main.c` and the bridge share one init/teardown is exactly the right structural fix (item C in the earlier review), and it also resolved the w64 archive name, signals/`alarm`, `interfaces_shutdown`, and the native `fyne-ui` target. 👍

One thing the refactor re-introduced, which I've just fixed on the branch:

**The #119 retry-slot regression came back — and this time in the production binary.**
`mercury_engine_init()` applied the connection config with `arq_set_retry_slots(cfg->data_retry_slots)` (the all-slots setter), which inflates CALL/ACCEPT retries to the DATA default (10). And `main.c` still ran its own (correct) config block *and then* called `mercury_engine_init()`, so the engine's call ran **last and won** → not just the Fyne UI, the standalone `mercury` binary too regressed to the ~90 s `ACCEPTING` linger after the caller gives up.

Pushed `c3cd9ee`:
- `mercury_engine.c`: `arq_set_retry_slots()` → `arq_set_data_retry_slots()` (data-plane retries no longer touch connection setup).
- `main.c`: removed the duplicated ARQ/tx-gain/busy config block — `mercury_engine_init()` already applies the identical sequence from the same `mcfg`. Keeping both is exactly what let the two copies drift.

Verified: clean build, full unit suite green, `TestMercuryARQTransfer`/`Control` integration green (CALL/ACCEPT back to the short default of 4, matching `mercuryv2`).

Remaining (minor, non-blocking): the `strncpy` truncation `-Wstringop-truncation` warnings in `mercury_engine.c` — same pattern used elsewhere; could add an explicit NUL to silence.

With `c3cd9ee` in, I don't see any remaining blockers from the review.

PR #126, opened 2026-07-19
freedv/HARQ: average combined LLRs by copy count (not raw-sum)

rafael2k:
## What

HARQ Chase combining previously **raw-summed** the retained LLRs across retransmissions (`llr_comb = llr_raw + harq_llr`). codec2's LDPC decoder is exact **sum-product** (`SumProduct`/`phi0`) — scale-sensitive — and its LLRs come from a fixed `EsNodB` (an uncalibrated scale). Summing N copies inflated the magnitude ~N×, over-driving the decoder out of its calibrated range and eroding combining gain on the large codes every Mercury ARQ mode uses (`H_256_768_22`, `H_16200_9720`).

**Fix:** track `harq_ncopies` and **average** — `llr_comb = (llr_raw + harq_llr) / (ncopies+1)` — keeping the magnitude at ~single-copy scale while the independent per-copy noise still averages down. N=1 is a pure no-op (single-shot path untouched; averaging only in the multi-copy fallback).

## Measured (deterministic A/B, same faded file, avg vs old raw-sum; delivered frames)

**DATAC15** (`H_256_768_22`, rate-1/3):

| SNR3k | single | HARQ avg | HARQ sum |
|---|---|---|---|
| −13.8 dB | 0 | 2 | 1 |
| −12.8 dB | 2 | 12 | 11 |
| −11.8 dB | 30 | 34 | 34 |
| −9.8 dB | 76 | 76 | 76 |

**DATAC17** (`H_16200_9720`, rate-0.6 — the large code) — the fix matters most here:

| SNR3k | single | HARQ avg | HARQ sum |
|---|---|---|---|
| −0.8 dB | 0 | **24** | 11 |
| +1.2 dB | 0 | **27** | 21 |
| +3.2 dB | 41 | 46 | 46 |
| +5.2 dB | 60/60 | 60 | 60 |

Averaging is **never worse and wins at the deep fringe**, with a large margin on the rate-0.6 `H_16200_9720` code (≈2× at −0.8 dB) — the regime OpenARQ's (N4FPV) reference benches independently flagged raw-sum as net-counterproductive. QAM16C2 shares that code and behaves the same class.

## Scope / testing
- 3 files: `modem/freedv/freedv_700.c`, `freedv_api.c`, `freedv_api_internal.h`.
- Build + full unit suite green; branched off `mercuryv2` @ 1.9.10.
- Isolated freedv-layer change — independent of the mfsk-arq work.

### DIFF
--- a/modem/freedv/freedv_700.c
+++ b/modem/freedv/freedv_700.c
@@ -266,6 +266,7 @@ void freedv_ofdm_data_open(struct freedv *f, struct freedv_advanced *adv) {
   f->harq_llr_nbits = 0;
   f->harq_enable = 0;
   f->harq_valid = 0;
+  f->harq_ncopies = 0;
 }
 
 /* speech or raw data, complex OFDM modulation out */
@@ -588,8 +589,18 @@ int freedv_comp_short_rx_ofdm(struct freedv *f, void *demod_in_8kHz,
         if (!crc_ok && f->harq_enable && f->harq_valid &&
             f->harq_llr_nbits == Npayloadbitsperpacket) {
           float llr_comb[Npayloadbitsperpacket];
+          /* Average (normalize by copy count) — do NOT raw-sum.  codec2 derives
+           * LLRs from a fixed EsNodB (an uncalibrated scale), so summing N
+           * copies inflates the magnitude ~Nx and over-drives the scale-
+           * sensitive sum-product (phi0) LDPC decoder, losing combining gain on
+           * the large codes (H_256_768_22, H_16200_9720) that every Mercury ARQ
+           * mode uses.  Averaging keeps the magnitude in the decoder's
+           * calibrated range while the independent per-copy noise still averages
+           * down (~3 dB per doubling).  harq_llr holds the sum of harq_ncopies
+           * prior copies; llr_raw is this copy, so divide by ncopies + 1. */
+          float norm = 1.0f / (float)(f->harq_ncopies + 1);
           for (int i = 0; i < Npayloadbitsperpacket; i++)
-            llr_comb[i] = llr_raw[i] + f->harq_llr[i];
+            llr_comb[i] = (llr_raw[i] + f->harq_llr[i]) * norm;
           ldpc_decode_frame(ldpc, &parityCheckCount, &iter, decoded_codeword,
                             llr_comb);
           memcpy(f->rx_payload_bits, decoded_codeword, Ndatabitsperpacket);
@@ -641,16 +652,20 @@ int freedv_comp_short_rx_ofdm(struct freedv *f, void *demod_in_8kHz,
         if (f->harq_enable) {
           if (crc_ok) {
             f->harq_valid = 0;
+            f->harq_ncopies = 0;
           } else {
-            /* Retain the running SUM of single-shot LLRs across retries so the
-             * next attempt's fallback combines all prior copies of this frame. */
+            /* Retain the running SUM of single-shot LLRs plus the copy count, so
+             * the next attempt's fallback averages all prior copies of this
+             * frame (see the averaging rationale at the combine site above). */
             if (f->harq_valid && f->harq_llr_nbits == Npayloadbitsperpacket) {
               for (int i = 0; i < Npayloadbitsperpacket; i++)
                 f->harq_llr[i] += llr_raw[i];
+              f->harq_ncopies++;
             } else {
               memcpy(f->harq_llr, llr_raw,
                      sizeof(float) * Npayloadbitsperpacket);
               f->harq_llr_nbits = Npayloadbitsperpacket;
+              f->harq_ncopies = 1;
             }
             f->harq_valid = 1;
           }
--- a/modem/freedv/freedv_api.c
+++ b/modem/freedv/freedv_api.c
@@ -1345,9 +1345,9 @@ void freedv_set_test_frames(struct freedv *f, int val) { f->test_frames = val; }
  * payload rather than a retransmission of the failed one. */
 void freedv_set_harq(struct freedv *f, int enable) {
   f->harq_enable = enable ? 1 : 0;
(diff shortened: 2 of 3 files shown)


kc1wzq:
It’s like Q65 right? Only issue is phase shift from the ionosphere.. could
ruin averages.


73s
Nicholas Charros, KC1WZQ

rafael2k:
> It’s like Q65 right? Only issue is phase shift from the ionosphere.. could ruin averages. 73s Nicholas Charros, KC1WZQ
> […](#)

Hi @kc1wzq, thanks for the alert.I'm running more tests on this to make sure we don't get worse. I get back soon.

rafael2k:
Good analogy — it's the Q65 family of "average repeated transmissions," and you're right that phase drift would wreck this **if it were coherent** combining. It isn't: we combine **post-demodulation LLRs** (soft bits), not complex symbols. Each retransmission is independently acquired and equalized — its own preamble + pilots resolve carrier frequency and phase *within* the burst — so only the real-valued bit reliabilities are averaged. Inter-transmission ionospheric phase shift / Doppler never accumulates into the average. It's non-coherent soft combining, phase-robust in the same spirit as Q65 itself (which accumulates non-coherently too). This PR only changes averaging vs raw-summing those LLRs, to keep the magnitude in codec2's sum-product decoder's calibrated range — orthogonal to the coherence question.

To show it rather than assert it, here's the same avg-vs-sum A/B on a **fading channel with Doppler/phase spread** — ITU-R Watterson "poor" — for DATAC17 (the rate-0.6 `H_16200_9720` code), same faded file decoded three ways, delivered/detected out of 40:

| SNR3k | single-shot | HARQ (avg, this PR) | HARQ (raw-sum, before) |
|---|---|---|---|
| +4.2 dB | 0/40 | **20/40** | 18/40 |
| +6.2 dB | 10/40 | 23/40 | 23/40 |

Even with the phase variation "poor" injects, combining still rescues 0 → 20 frames, and averaging is ≥ raw-sum. Phase drift doesn't ruin it precisely because the phase is resolved per burst *before* the LLRs are formed; the average is over reliabilities, not over a rotating constellation.

rafael2k:
> Good analogy — it's the Q65 family of "average repeated transmissions," and you're right that phase drift would wreck this **if it were coherent** combining. It isn't: we combine **post-demodulation LLRs** (soft bits), not complex symbols. Each retransmission is independently acquired and equalized — its own preamble + pilots resolve carrier frequency and phase _within_ the burst — so only the real-valued bit reliabilities are averaged. Inter-transmission ionospheric phase shift / Doppler never accumulates into the average. It's non-coherent soft combining, phase-robust in the same spirit as Q65 itself (which accumulates non-coherently too). This PR only changes averaging vs raw-summing those LLRs, to keep the magnitude in codec2's sum-product decoder's calibrated range — orthogonal to the coherence question.
> 
> To show it rather than assert it, here's the same avg-vs-sum A/B on a **fading channel with Doppler/phase spread** — ITU-R Watterson "poor" — for DATAC17 (the rate-0.6 `H_16200_9720` code), same faded file decoded three ways, delivered/detected out of 40:
> SNR3k 	single-shot 	HARQ (avg, this PR) 	HARQ (raw-sum, before)
> +4.2 dB 	0/40 	**20/40** 	18/40
> +6.2 dB 	10/40 	23/40 	23/40
> 
> Even with the phase variation "poor" injects, combining still rescues 0 → 20 frames, and averaging is ≥ raw-sum. Phase drift doesn't ruin it precisely because the phase is resolved per burst _before_ the LLRs are formed; the average is over reliabilities, not over a rotating constellation.

It seems from testing data things are ok. We'll carry some OTA validation too.

kc1wzq:
The issue I’m seeing is even though the preamble acquired the phase, the
phase can be quick to change. It’s really unpredictable on HF. I see this
working on VHF or even EME where phase is stable though!


73s
Nicholas Charros, KC1WZQ

rafael2k:
Thanks Nicholas — you're pointing at exactly the right failure mode, and you're right that fast HF phase is brutal. The key distinction is *what* breaks under it: **acquisition/tracking**, not the LLR combining.

Two things happen within a burst: (1) these OFDM modes track phase **continuously via interspersed pilots**, not frozen at the preamble — DATAC15/16/17 carry a pilot row every few symbols, so the equalizer re-estimates channel phase many times across the burst. (2) The LLRs are formed **after** that per-burst equalization, so combining averages *reliabilities*, never a rotating constellation. So the real question is just: can the pilots track the Doppler?

Measured it (DATAC17, `H_16200_9720`, Watterson):

| channel (Doppler) | single-shot | HARQ | acquisition |
|---|---|---|---|
| moderate (~0.5 Hz) +7 dB | 22/40 | 27/40 | fine |
| poor (~1 Hz, deep) +4 dB | 0/40 | 20/40 | fine → big rescue |
| flutter (fast/auroral) +11 dB | 0/40 | 0/40 | **only ~5/40 detected** |

On moderate/poor (typical HF), the pilots track the phase within the burst, acquisition works, and combining helps. On **flutter**, the mode barely *acquires* (~5 of 40 frames sync) and decodes 0 — and HARQ can't help because there's nothing synced to combine. Importantly it doesn't *hurt* either (0/5 both ways). That's a fundamental acquisition limit for any post-sync scheme — the same thing OpenARQ's implementation notes call out (acquisition, not FEC/combining, is the low-SNR limiter; combining gives ~0 dB to acquisition). At that point you'd drop to a robust narrowband mode / the weak-signal floor — and even then it's an *acquisition* floor, not a combining one.

Net: VHF/EME (stable phase) — ideal, agreed; typical fading HF — pilots track it and combining helps (shown); fast-flutter HF — acquisition-limited regardless of how you combine. And this PR is orthogonal to all of it: it only rescales the combined LLRs, and it's a no-op when acquisition fails (flutter, 0/5 either way).

PR #125, opened 2026-07-18
arq: MFSK weak-signal backend + delivery-driven stop-and-wait rethink

rafael2k:
## What

A ground-up rethink of the ARQ **data plane** plus a second modem backend, rebased onto the `v1.9.10` base.

**Modem abstraction (the intended v2 design):**
- `modem_backend_t` vtable — FreeDV is backend 1, and a new **MFSK weak-signal backend** (backend 2, `MERCURY_MODE_MFSK`).
- Welch-Costas **pattern ACK** — ~0.64 s and survives ~10 dB deeper than a DATAC16 ACK.

**Data-flow FSM — simpler, faster:**
- 21 → 11 states; data flow 15 → 5 (`IDLE_ISS → DATA_TX → WAIT_ACK`, `IDLE_IRS → ACK_TX`), piggyback turn handoff.
- **Delivery-driven mode ladder, no SNR:** starts at the most robust rung (MFSK) and climbs on clean deliveries, steps down on any retry.
- Single retained `tx_frame` (no window/restage); removes OLLA/SNR interlock, MODE/TURN/KEEPALIVE negotiation, and the restage buffer.
- Designed for **asymmetric links** (per-direction adaptation + robust pattern ACK on the reverse path).

## Validation (deterministic instruments — full green)

- **Unit + sim suite: 172 tests / 15 binaries, all pass**; native `mercury` builds clean on the rebased tree.
- **Two-FSM sim (`test_arq_sim`, 17 scenarios):** clean, 20% loss, **fade-cliff → MFSK downgrade**, **asymmetric strong-forward**, **bidirectional piggyback turn**, **lost-ACK idempotency**, + flat/fading fuzz.
- **Ladder (5)** and **pattern-ACK DSP detection (3, incl. 0 false-alarm on noise)**.
- **`ab_bench` throughput/integrity matrix (18 runs, clean/awgn/cliff/nvis): integrity OK on 100%**; ladder climbs to top on clean links and drops to the MFSK floor on harsh channels.

## ⚠️ Merge gate

**Do not merge until on-air validation** (task 31 — São Roque ↔ Belo Horizonte, sbitx `-S`). The sims are the pre-OTA instrument; the real gate is Pedro's OTA run. Opening now for review.

Rebased onto `v1.9.10` (`c1e8f56`); 13 commits, no conflicts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #127, opened 2026-07-21
freedv/HARQ: gate combined-decode acceptance on parity count, not CRC16 alone

spinkham:
The HARQ Chase fallback re-decodes the averaged LLRs and then accepts the
result on freedv_check_crc16_unpacked() alone; parityCheckCount is computed
but only printed in the verbose path. A combine that has NOT converged still
emits a codeword, and that codeword still has a ~2^-16 chance of passing CRC16
by itself. On the single-shot path that stray pass is just a rare undetected
bit error. On the combine path it is reached on purpose at the fade cliff --
roughly half the combines fail to converge there by design -- so the roll is
taken far more often than 2^-16-per-frame suggests. Today burst_frames==1
everywhere bounds the blast radius (a same-codeword combine, one DATA frame
per session), but the moment burst_frames>1 lets the RX admit more than one
DATA frame per session, a CRC-passing non-converged combine parses in-window
and is delivered as new data instead of being retransmitted -- silent
corruption, not a retry.

Fix: require the LDPC parity-check count to clear a wide margin (>= 90% of
NumberParityBits, via the new ldpc_harq_combine_parity_ok()) in ADDITION to
CRC16 before adopting a combined decode. A converged decode satisfies
essentially all of its mother-code parity checks; a non-converged one sits
near the ~50% binomial floor, so 90% cleanly separates the two populations.
Deliberately not a strict all-checks gate: at the marginal combining point
residual errors can live in parity bits, so a genuine decode need not reach
100% and a strict gate would reject it. The single-shot decode is untouched,
so clean-channel goodput is unchanged; RX-only, no wire-format change.

Measured on Mercury's own H_256_768_22 (DATAC15) decoder + CRC16 with a
fixed-seed AWGN Chase model: 3,000,000 sub-threshold trials produced 47
CRC16-passing NON-converged decodes -- the silent forgeries CRC-only
acceptance would deliver as data (rate ~2^-16, as expected). ZERO of the 47
cleared the 90% parity gate (worst forgery 402/512, gate 460/512). Separately,
200,000 genuine near-threshold decodes ALL cleared the gate (min 507/512), so
it never false-rejects a real decode. The threshold sits cleanly between the
two populations: 402 (worst forgery) < 460 (gate) < 507 (worst genuine).
Corroborated by OpenARQ's reference implementation at the same decode point.

Test (tests/modem/test_freedv_harq.c) drives Mercury's own DATAC15 decoder
(run_ldpc_decoder on H_256_768_22) with a fixed-seed AWGN model and asserts
the separation: every genuine correct decode clears the gate (no false-reject),
a non-converged garbage decode never does (forgery blocked), and the threshold
is exact at the boundary. ldpc_harq_combine_parity_ok is the single source of
the threshold shared by the decoder and the test.

Complements d356833 (the LLR-averaging half of the same combining path):
d356833 fixed the combined LLR *scale*, this fixes its *acceptance*.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

---
Opened via Claude Code (Opus 4.8) working with N4FPV; provenance in the commit trailer.

### DIFF
--- a/modem/freedv/freedv_700.c
+++ b/modem/freedv/freedv_700.c
@@ -603,9 +603,31 @@ int freedv_comp_short_rx_ofdm(struct freedv *f, void *demod_in_8kHz,
             llr_comb[i] = (llr_raw[i] + f->harq_llr[i]) * norm;
           ldpc_decode_frame(ldpc, &parityCheckCount, &iter, decoded_codeword,
                             llr_comb);
-          memcpy(f->rx_payload_bits, decoded_codeword, Ndatabitsperpacket);
-          crc_ok = freedv_check_crc16_unpacked(f->rx_payload_bits,
-                                               Ndatabitsperpacket);
+          /* Integrity gate on the COMBINED decode — do NOT accept on CRC16
+           * alone.  A Chase combine that has NOT converged still emits a
+           * codeword, and that codeword still has a ~2^-16 chance of passing
+           * CRC16 by itself.  On the single-shot path that stray pass is just a
+           * rare undetected bit-error; on the combine path it is reached on
+           * PURPOSE at the fade cliff (roughly half the combines fail to
+           * converge there by design), and once burst_frames>1 lets the RX
+           * admit more than one DATA frame per session such a forgery parses
+           * in-window and is delivered as new data rather than retransmitted.
+           * A converged LDPC decode satisfies essentially all of its
+           * mother-code parity checks; a non-converged one sits near the ~50%
+           * binomial floor.  So require the parity-check count to clear a wide
+           * margin (90% of NumberParityBits) IN ADDITION to CRC16 before
+           * adopting a combined result.  Deliberately NOT a strict all-checks
+           * gate: at the marginal combining point residual errors can live in
+           * parity bits, so a genuine decode need not reach 100% (measured e.g.
+           * 4094/4096 on the OpenARQ reference) — 90% cleanly separates the
+           * converged and garbage populations.  Single-shot decodes above are
+           * untouched, so clean-channel goodput is unchanged. */
+          if (ldpc_harq_combine_parity_ok(parityCheckCount,
+                                          ldpc->NumberParityBits)) {
+            memcpy(f->rx_payload_bits, decoded_codeword, Ndatabitsperpacket);
+            crc_ok = freedv_check_crc16_unpacked(f->rx_payload_bits,
+                                                 Ndatabitsperpacket);
+          }
         }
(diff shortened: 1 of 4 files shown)


spinkham:
Another option: Someone else pushed an independent fix for the same issue today: https://github.com/xmutantson/mercury/commit/ffcb1f457d2d65398227cabdeaac0165abbcef57

They way their fix is designed has a slightly tighter gate, which since there is no stream CRC in Mercury, it may be a better solution for you to adopt instead of this one. I'd recommend you evaluate both.

rafael2k (approved):
Reviewed + validated locally — approving.

**The bug is real.** In `freedv_comp_short_rx_ofdm()` the *combined* decode was accepted on `freedv_check_crc16_unpacked()` alone (`parityCheckCount` was computed but only used in the verbose print). At the fade cliff the combine path is entered on purpose and ~half the combines haven't converged, so the ~2⁻¹⁶ stray-CRC roll is taken far more often than per-frame odds suggest. Today `burst_frames==1` bounds the blast radius, but it turns into silent corruption the moment `burst_frames>1` admits >1 DATA frame/session — i.e. exactly when we resume multi-frame bursts.

**The fix is sound and minimal:** the 90% parity-count gate is applied only to the combine path (single-shot untouched → clean-channel goodput unchanged), it's *additive* to CRC16, and 90% cleanly separates converged (~all checks) from non-converged (~50% binomial floor) while staying below 100% so genuine-but-marginal combines aren't rejected. `ldpc_harq_combine_parity_ok()` as a shared inline (RX + test single source of truth) is nice.

**Validation:**
- `test_freedv_harq` builds and passes 3/3 (boundary; correct-decode-always-clears; non-converged-never-clears), driving the real DATAC15 `H_256_768_22` decoder.
- **Goodput-neutrality A/B** — decoded identical CCIR-poor faded DATAC17 files (across the whole HARQ-active cliff) with gate-OFF (mercuryv2) vs gate-ON, both `--harq`. Delivered counts are **identical at every SNR** (18/38, 14/38, 12/37, 7/32, 6/31, 3/24, 0/10 — MATCH throughout), confirming the gate rejects zero genuine combined deliveries. (Channel: OpenARQ skywave/hfchan, noise axis calibrated to codec2 `ch`.)

Nice catch — forward-defense on precisely the path our roadmap lights up. Note: the parallel `mfsk-arq-integration` branch (#125) also carries this HARQ code, so it'll need to pick up this gate on its next rebase onto mercuryv2.

PR #129, opened 2026-07-22
audio: -x sock — a virtual-clock lockstep backend for device-free testing

spinkham:
## What this adds

A new audio backend, `-x sock`, that connects mercury to a channel simulator over an AF_UNIX socket carrying framed audio blocks. Unlike `-x fifo`, the transport also carries time: each simulator frame includes `virtual_now_ms`, and mercury's ARQ timers advance on that clock instead of wall time.

The result is bench testing with no sound card, no real-time pacing, and no OS-dependent timing:

- Two mercury instances + a simulator complete full ARQ sessions (connect, mode ladder, 4 KB transfers, half duplex with real PTT) about 10x faster than real time.
- Runs behave identically on Linux and macOS (Apple Silicon). We measured 58.9 vs 57.7 B/s real-time-equivalent goodput on the same cell across the two platforms.
- No starvation flakes: a slow or loaded host slows the run down instead of breaking it, because nothing depends on keeping up with a wall clock.

Everything is inert unless `-x sock` is selected: the virtual clock is off by default, and `time_now_ms()` is exactly `hermes_uptime_ms()` when it is off, so all existing backends and platforms are behavior-identical.

## How it works

- `common/virtual_clock.{c,h}`: a process-wide time source. `time_now_ms()` returns wall uptime normally, or the transport-provided time when the sock backend enabled it at engine init.
- `audioio/audioio.c`: the lockstep transport thread. Per block: receive the sim frame, deposit samples in the capture ring, advance the virtual clock, wake the ARQ event loop, reply with TX samples plus the real PTT state. It backpressures the simulator so signal time can never outrun the demodulator.
- `audioio/sock_wire.h`: the wire codec (little-endian, u32 length prefix, seq/virtual_now_ms/n headers), pinned byte-exact by `tests/audioio/test_sock_wire.c`.
- `datalink_arq/`: all FSM clock reads routed through `time_now_ms()` (mechanical, identity when virtual is off), plus `arq_notify_virtual_time()` so the event loop re-reads time when the transport advances it.
- `modem.c`: under a virtual clock only, PTT spans the burst in signal time, the mode-switch debounce counts in signal time, and the TX-to-RX capture flush is skipped (its "stale audio" premise doesn't hold when deadlines only advance with delivered blocks).

## The wire contract

One frame per audio block, 8 kHz mono i16 (mercury's native rate, no resampling):

```
u32 len                          bytes after this field
sim -> station:  u64 seq | u64 virtual_now_ms | u16 n | n i16 samples
station -> sim:  u64 seq | u8 ptt | u16 n | n i16 samples
```

The simulator sends the RX block first, then waits for the station's reply (lockstep barrier). A reference simulator harness lives in the skywave bench (openarq-project/skywave, `sock_frames.py` / `channel_sim.py`), but the contract above is self-contained and small enough to reimplement in ~50 lines for standalone testing.

## Motivation

We benchmark mercury alongside other HF modems and wanted CI-grade, cross-platform integration runs. The `-x fifo` path needs the harness to synthesize an 8 kHz wall clock, which races the modem on fast hosts (reproducibly stalls on an Apple Silicon Mac). Carrying the clock in the transport removes the race entirely. This work also surfaced the macOS thread-stack crash fixed in #128, which this branch is based on (the first commit here is that fix; it drops out once #128 merges).

## Testing

- `tests/`: new `test_sock_wire` (byte-exact frame layout) and `test_virtual_clock`; whole suite green on Linux and macOS (the suite also gained the missing `virtual_clock.c` link lines for the arq test binaries).
- End-to-end: repeated 4096/4096-byte transfers, full duplex and half duplex with in-band PTT, clean and AWGN cells, on Linux x86-64 and macOS arm64.

Happy to split, rename, or adjust the wire format if you'd prefer different conventions. If the backend itself isn't of interest, #128 stands alone.

https://claude.ai/code/session_01Az9HBCH2P8e5f2A4G5LPTn


rafael2k:
Thanks!

PR #130, opened 2026-07-22
Store config, logs in install dir on Windows and fix audio device persistence on restart

pedromessetti:
1. **Consolidate config and log paths to the installation directory**
On Windows, the GUI now stores `mercury.ini`, `mercury_engine.log`, and `ui.log` all in the same folder as the program executable (`C:\Program Files\MercuryHF\`), instead of scattering config to `%LOCALAPPDATA%\MercuryHF\.` The first-run copy-to-per-user-dir logic is removed.
2. **Grant write permissions in the installer**
Added `[Dirs]` with `Permissions: users-modify` on `{app}` and `Permissions: users-modify` on `mercury.ini` in the Inno Setup script, so non-admin users can write config changes and create log files in the install directory.
3. **Fix audio device selection not persisting across restarts**
`apply_audio_defaults()` was unconditionally zeroing the device ID strings for DirectSound, WASAPI, and PulseAudio backends after loading them from `mercury.ini`. This caused saved soundcard selections to be lost on every restart — the UI showed defaults and no audio was received until devices were manually re-selected. Removed the destructive clearing; empty string already means "default device" for these backends.
4. **Switch Windows default audio system to WASAPI**
Changed the installer's default sound_system from dsound to wasapi.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -355,7 +355,7 @@ fyne-ui-windows: libmercury_core_w64.a
 windows-installer: fyne-ui-windows
 	cp mercury.ini.example $(WINDOWS_INSTALLER_DIR)/mercury.ini
 	sed -i 's/ui_enabled = false/ui_enabled = true/g' $(WINDOWS_INSTALLER_DIR)/mercury.ini
-	sed -i 's/sound_system = auto/sound_system = dsound/g' $(WINDOWS_INSTALLER_DIR)/mercury.ini
+	sed -i 's/sound_system = auto/sound_system = wasapi/g' $(WINDOWS_INSTALLER_DIR)/mercury.ini
 	if ls $(HAMLIB_W64_DIR)/bin/*.dll >/dev/null 2>&1; then \
 		cp $(HAMLIB_W64_DIR)/bin/*.dll $(WINDOWS_INSTALLER_DIR)/; \
 	fi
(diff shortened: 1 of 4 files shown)


common/mercury_engine.c
```diff
@@ -55,8 +55,6 @@ static int apply_audio_defaults(int audio_system, char *input_dev, size_t in_siz
     case AUDIO_SUBSYSTEM_PULSE:
     case AUDIO_SUBSYSTEM_WASAPI:
     case AUDIO_SUBSYSTEM_DSOUND:
-        input_dev[0] = '\0';
```
rafael2k, line 58: Why remove?

Makefile
```diff
@@ -355,7 +355,7 @@ fyne-ui-windows: libmercury_core_w64.a
 windows-installer: fyne-ui-windows
 	cp mercury.ini.example $(WINDOWS_INSTALLER_DIR)/mercury.ini
 	sed -i 's/ui_enabled = false/ui_enabled = true/g' $(WINDOWS_INSTALLER_DIR)/mercury.ini
-	sed -i 's/sound_system = auto/sound_system = dsound/g' $(WINDOWS_INSTALLER_DIR)/mercury.ini
+	sed -i 's/sound_system = auto/sound_system = wasapi/g' $(WINDOWS_INSTALLER_DIR)/mercury.ini
```
rafael2k, line 358: better is the auto adopt wasapi by default

rafael2k (changes requested):
I did not understand some details.

rafael2k:
LGTM

PR #133, opened 2026-07-27
windows: sign the installer and its payloads; ship the console binary

rafael2k:
Makes the Windows release fully signed — the installer and everything inside it — and improves the signing tooling's error reporting.

## Installer signing

Previously the installer path signed only the finished `Setup.exe`, while the payload staged into it was unsigned. The result was a signed installer that installs unsigned programs: the publisher shows on the download, then not when the program is run.

New `windows-installer-signed` target signs the payloads first, then builds the installer, then signs it — in a single signing session.

## Console binary in the installer

`mercury.exe` (console/TNC build) is now packed and signed alongside the GUI, so Windows users have it without downloading the ZIP separately. It is needed for headless, TNC and scripted use, and for diagnosing problems.

Kept out of a first-time user's way: no desktop icon, Start Menu entry under **Advanced**, and the shortcut opens a command prompt showing the usage text rather than launching the modem. Installed unconditionally rather than as an optional component — one less choice in the wizard.

Also removes `run_mercury.bat`: nothing referenced it (all shortcuts point directly at the GUI executable), and it added a third ambiguous file to the install folder.

## Signing tooling

- `code-signing/sign-diag.sh`: reports the local signing prerequisites; touches no remote service. Prints no secrets.
- Missing prerequisites are now detected up front and named, instead of surfacing later as a generic failure.
- Failures capture diagnostic output for the automated login step, which is otherwise opaque.
- Three robustness fixes in session teardown and display handling, to avoid interfering with the operator's own desktop session.

## Validation

- Full chain exercised end to end: both payload executables and the installer signed, all verifying, with the payloads signed before the installer is built.
- Clean build; test suite green (201 tests, 0 failures).
- Rebased onto current `mercuryv2`.

## Notes

- Requires Inno Setup for the installer step; set `ISCC` and (on Linux) `ISCC_RUN`. Left unset, the target signs the payloads and prints the remaining commands.
- `installer.iss` carries its own version string; keep it in step with `MERCURY_VERSION`.
- See `docs/WINDOWS-SIGNING.md` for setup and troubleshooting.


pedromessetti:
LGTM

PR #146, opened 2026-07-31
arq/fsm: defer LISTEN OFF during grace period after PENDING

pedromessetti:
## Problem

Scanning hosts (BPQ32) send LISTEN OFF at dwell expiry. When a CALL frame decodes near the end of a dwell window, the LISTEN OFF crosses PENDING on the wire: the host has already issued the channel release before processing the incoming-call notification.

In v1.9.11 the `aafd612` fix began honouring LISTEN OFF in CALLING/ACCEPTING (correct per the VARA spec — the host wants the radio back), but this exposed the race: the handshake was cancelled, the host moved to the next channel, and the ACCEPT reply went out on the wrong frequency.

**Reproduction** (Richard KO0OOO, 7 s dwell, 4-frequency scan):
- Connect request arriving within the first 4 s of dwell → scan stops ✓
- Connect request arriving after 4 s → scan does NOT stop on that channel; stops on the next channel and starts tx'ing there ✗
- v1.9.10 gateway + v1.9.10 client worked throughout the full dwell window

## Fix

Add `ARQ_LISTEN_OFF_GRACE_MS` (2000 ms). A LISTEN OFF that arrives within 2 s of entering CALLING or ACCEPTING is deferred rather than acted on immediately. The deadline is re-armed so TIMER_RETRY fires at grace expiry, at which point the deferred event is honoured. 

This gives the host time to process PENDING and cancel its own dwell timer. LISTEN OFF after the grace window is still acted on immediately (genuine channel release via INTERLOCK or operator request).

## Changes

| File | Change |
|------|--------|
| `arq_protocol.h` | Add `ARQ_LISTEN_OFF_GRACE_MS = 2000` |
| `arq_fsm.h` | Add `deferred_listen_off` flag to session struct |
| `arq_fsm.c` | Defer LISTEN OFF in CALLING/ACCEPTING within grace; honour on TIMER_RETRY after expiry |
| `test_arq_fsm.c` | Update existing tests for grace advance; add 2 new tests for deferred-vs-immediate behaviour |

## Tests

```
31 Tests 0 Failures 0 Ignored OK
```

Two new tests:
- `test_listen_off_deferred_within_grace_accepting` — LISTEN OFF within 2 s defers; TIMER_RETRY after grace honours it
- `test_listen_off_deferred_within_grace_calling` — same for CALLING state

Existing `test_listen_off_drops_pending_accept` and `test_listen_off_drops_outgoing_call` updated to advance mock time past the grace period before dispatching LISTEN OFF (maintains the original intent: after the grace window, the event is immediate).

## Stacking

This PR targets `feat-busy-default-on` (#144). The two fixes are complementary:
- #144 enables BUSY by default for ~0.3 s early carrier detection (helps if the host understands BUSY)
- This PR adds the LISTEN OFF grace period to prevent the race regardless of BUSY support

### DIFF
--- a/datalink_arq/arq_fsm.c
+++ b/datalink_arq/arq_fsm.c
@@ -156,6 +156,7 @@ static void sess_enter(arq_session_t *sess, arq_conn_state_t new_state,
     sess->state_enter_ms = time_now_ms();
     sess->deadline_ms    = deadline_ms;
     sess->deadline_event = deadline_event;
+    sess->deferred_listen_off = false;  /* stale across state transitions */
     if (new_state != ARQ_CONN_CONNECTED)
     {
         sess->pending_connect_confirm = false;
@@ -1175,6 +1176,14 @@ static void fsm_calling(arq_session_t *sess, const arq_event_t *ev)
         break;
 
     case ARQ_EV_TIMER_RETRY:
+        if (sess->deferred_listen_off &&
+            time_now_ms() - sess->state_enter_ms >= ARQ_LISTEN_OFF_GRACE_MS)
+        {
+            sess->deferred_listen_off = false;
+            if (g_cbs.notify_disconnected) g_cbs.notify_disconnected(false);
+            enter_idle_after_call(sess);
+            break;
+        }
         if (sess->tx_retries_left > 0)
         {
             sess->tx_retries_left--;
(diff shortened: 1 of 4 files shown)


rafael2k:
Rebased onto `mercuryv2` (#144 is merged, so the old base is gone) and pushed two corrections. The race this fixes is real; the deferral just needed to not lose the request.

**1. A deferred LISTEN OFF was dropped whenever the handshake succeeded.** `sess_enter()` clears the flag on *every* transition, so: host asserts the interlock inside the grace → we defer → the call completes → `ACCEPTING → CONNECTED` wipes the flag → we settle into a session on a channel we were told to release. The interlock was honoured only when the call failed anyway, which is the opposite of its purpose — and it is exactly Gary's original report, reappearing on the success path.

The flag now survives into CONNECTED and is acted on there; it is cleared only when the session is already going idle, where there is nothing left to release.

**2. CALLING no longer defers.** The rationale for the grace is that the host needs a moment to process PENDING — but PENDING announces an *incoming* call and is never sent while we are the caller. There is no race to protect there, and delaying only keeps a radio the operator asked for. The now-unreachable deferral branch in its `TIMER_RETRY` is removed with it.

**Tests:** the CALLING test now asserts immediate release, and `test_deferred_listen_off_survives_connect` covers defer-then-connect — the case that decides whether the interlock still means anything, and the one the original tests did not reach. Full suite green.

**Unchanged:** `ARQ_LISTEN_OFF_GRACE_MS` keeps its 2 s and still applies to ACCEPTING, the state the reported race actually occurs in. No timing constant was touched.

Worth confirming with Richard once released, since his 7 s dwell is what produced the race.

PR #149, opened 2026-08-02
Add audio device matching by name

ptsmonteiro:
To make life easier on setups where the audio device id changes frequently, we can now have matching by name.

```
% ./mercury -z                           
Rhizomatica Mercury Version 1.9.11 (git cc5bbc7e)
playback devices:
device: name: 'USB Audio CODEC'  id: '154'  default: (null)
device: name: 'BlackHole 16ch'  id: '86'  default: (null)
device: name: 'BlackHole 2ch'  id: '107'  default: (null)
device: name: 'MacBook Air Speakers'  id: '142'  default: (null)
device: name: 'Microsoft Teams Audio'  id: '124'  default: (null)
device: name: 'MO BH 2ch'  id: '59'  default: (null)
device: name: 'MO BH 16ch'  id: '63'  default: (null)
capture devices:
device: name: 'IPhone 12 Pro Microphone'  id: '162'  default: (null)
device: name: 'USB Audio CODEC'  id: '160'  default: (null)
device: name: 'BlackHole 16ch'  id: '86'  default: (null)
device: name: 'BlackHole 2ch'  id: '107'  default: (null)
device: name: 'MacBook Air Microphone'  id: '149'  default: (null)
device: name: 'Microsoft Teams Audio'  id: '124'  default: (null)
```

```
% ./mercury -i 'USB audio' -o 'usb audio'
Rhizomatica Mercury Version 1.9.11 (git cc5bbc7e)
(...)
10:50:36.543 [+0.098s] [INF] [audio-cap] device 'USB audio' matched by name to 'USB Audio CODEC' (id=160)
10:50:36.544 [+0.099s] [INF] [audio-play] device 'usb audio' matched by name to 'USB Audio CODEC' (id=154)
(...)
10:50:36.546 [+0.101s] [INF] [audio-play] I/O playback (154 'USB Audio CODEC') float32 / 48000Hz / 2ch / 40ms buffer
10:50:36.551 [+0.107s] [INF] [audio-cap] I/O capture (160 'USB Audio CODEC') float32 / 48000Hz / 2ch / 40ms buffer
(...)
```
```
% ./mercury -i 160 -o 154                
Rhizomatica Mercury Version 1.9.11 (git cc5bbc7e)
(...)
10:53:17.320 [+0.101s] [INF] [audio-play] I/O playback (154 'USB Audio CODEC') float32 / 48000Hz / 2ch / 40ms buffer
10:53:17.321 [+0.101s] [INF] [audio-cap] I/O capture (160 'USB Audio CODEC') float32 / 48000Hz / 2ch / 40ms buffer
(...)
```

I only tested it on macOS. Should work equally well on other platforms.

### DIFF
--- a/audioio/audioio.c
+++ b/audioio/audioio.c
@@ -13,6 +13,8 @@
 #include <limits.h>
 #include <errno.h>
 #include <signal.h>
+#include <string.h>
+#include <ffbase/stringz.h>
 #include "os_interop.h"
 #include <ffaudio/audio.h>
 #include "std.h"
@@ -82,6 +84,8 @@ static char s_capture_dev[256];
 static char s_playback_dev[256];
 static int s_buffers_initialized = 0;
 static volatile bool audio_shutdown_ = false;  // local stop flag for audio threads
+
+static void format_device_display(int audio_subsys, int mode, const char *id, char *out, size_t outsz);
 
 #define NULL_AUDIO_PERIOD_MS 20
 #define NULL_AUDIO_SAMPLES_PER_PERIOD 160
@@ -851,8 +855,12 @@ void *radio_playback_thread(void *device_ptr)
         goto cleanup_play;
     }
 
+    char play_dev_display[96];
+    format_device_display(audio_subsystem, FFAUDIO_DEV_PLAYBACK,
+                           device_ptr ? (const char *)device_ptr : NULL,
+                           play_dev_display, sizeof(play_dev_display));
     HLOGI("audio-play", "I/O playback (%s) %s / %dHz / %dch / %dms buffer",
-          device_ptr ? (const char *)device_ptr : "default",
+          play_dev_display,
           cfg->format == FFAUDIO_F_FLOAT32 ? "float32" :
           cfg->format == FFAUDIO_F_INT32   ? "int32"   :
           cfg->format == FFAUDIO_F_INT24_4 ? "int24in32" :
@@ -1195,8 +1203,12 @@ void *radio_capture_thread(void *device_ptr)
         goto cleanup_cap;
     }
 
+    char cap_dev_display[96];
+    format_device_display(audio_subsystem, FFAUDIO_DEV_CAPTURE,
+                           device_ptr ? (const char *)device_ptr : NULL,
+                           cap_dev_display, sizeof(cap_dev_display));
     HLOGI("audio-cap", "I/O capture (%s) %s / %dHz / %dch / %dms buffer",
-          device_ptr ? (const char *)device_ptr : "default",
+          cap_dev_display,
           cfg->format == FFAUDIO_F_FLOAT32 ? "float32" :
           cfg->format == FFAUDIO_F_INT32   ? "int32"   :
           cfg->format == FFAUDIO_F_INT24_4 ? "int24in32" :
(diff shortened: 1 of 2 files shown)


ptsmonteiro:
This is failing the windows build but apparently has nothing to do with the pull request:

```
Fetching the repository
  /usr/bin/git -c protocol.version=2 fetch --prune --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/*
  Error: fatal: unable to access 'https://github.com/Rhizomatica/mercury/': Problem with the SSL CA cert (path? access rights?)
  The process '/usr/bin/git' failed with exit code 128
```

rafael2k:
Thanks — useful feature, and the fallback ordering (exact id → exact name → unambiguous substring) is the right shape. I tested it on Linux against current `mercuryv2` and found two problems. Both are fixed on branch `fix-149-name-match`, so this can go in with a rebase + those two commits folded in.

## 1. Blocker: self-deadlock against #151 (merged after you opened this)

#151 added `s_pulse_lock` to serialise PulseAudio global state. `audioio_restart()` holds it across the whole uninit → init cycle, and the new `resolve_device_string()` calls land **inside** that region. They reach `get_soundcard_list()`, which takes the same **non-recursive** mutex.

Any audio-config change from the UI under `-x pulse` therefore hangs mercury permanently — and it hangs *after* `audioio_stop_threads()`, so audio is already down and never comes back. Reproduced here:

```
[INF] [audio-restart] stopping audio threads...      <- never reaches "restarted"
```

```
#3  ___pthread_mutex_lock (mutex=<s_pulse_lock>)
#4  get_soundcard_list (audio_system=1, mode=1, ...)
#5  resolve_device_string (...)
#6  audioio_restart (...)
#7  ws_command_handler (ui_communication.c:109)
#8  ws_handle_message (mercury_websocket.c:297)
```

Fix used: split the enumerator into `get_soundcard_list_int(..., bool pulse_lock_held)`, with the public `get_soundcard_list()` as a thin wrapper, and let the one caller that already owns the lock skip it. I preferred that to making the mutex recursive so the lock structure stays explicit. The invariant holds in both directions — whenever resolution would touch `s_pulse_lock` (subsystem is PULSE), `audioio_restart` has already taken it, because a *new* subsystem of PULSE is part of its lock condition.

After the fix the restart completes in ~10 ms and a subsequent client is still served (i.e. the lock was genuinely released).

## 2. Exact-name matching silently picks the first duplicate

```c
if (ffsz_ieq(buf, names[i])) { match = i; matches = 1; break; }
```

Stopping at the first hit and hard-coding `matches = 1` means identical display names never trigger the ambiguity warning — the substring path counts properly, the exact path doesn't.

This isn't exotic. On this host **both** PulseAudio playback devices are named `Built-in Audio Analog Stereo` (the PCI card and the snd-aloop one), so:

```
[INF] [audio-cap]  device 'Built-in Audio Analog Stereo' matched by name to
                   'Built-in Audio Analog Stereo' (id=alsa_input.pci-0000_00_1f.3.analog-stereo)
```

It silently took the PCI card with no warning. Two of the same USB codec — a very common ham setup — collide the same way. Given the feature exists precisely because ids move around, enumeration order isn't a stable tiebreak either, so this can bind the wrong radio.

Fixed by counting all exact matches and warning like the substring path does:

```
[WRN] [audio-cap] device name 'Built-in Audio Analog Stereo' is ambiguous (2 matches) -- use an exact id from -z instead
```

## Verified after both fixes

- unique substring resolves: `-i 'Loopback'` → `plughw:2,0 'Loopback Loopback PCM'`
- exact id passed through untouched (no name-matching path taken)
- duplicate names warn instead of guessing
- UI audio-config restart completes, no deadlock
- build clean, unit suite green

## Minor, not blocking

Each resolve and each startup log line runs a full enumeration, so a start now costs up to four device enumerations. Cheap on ALSA, a few server round trips on Pulse. Fine as-is, just worth knowing on slow Pi setups.

Happy to open the fixes as a PR into your branch, or you can cherry-pick `fix-149-name-match` — whichever you prefer.

rafael2k:
Thanks for this @ptsmonteiro. I'm reviewing.

ptsmonteiro:
Thanks @rafael2k.
I did a rebase from the current main mercuryv2 and it seems it can merge cleanly. If I understand correctly it will incorporate the fixes from fix-149-name-match (which was already merged).
Did I miss anything?

rafael2k:
Thanks for this @ptsmonteiro — and nice catch on the enumeration deadlock.

Quick heads-up on why this is now showing as conflicting: the two PulseAudio mutex commits (`audioio: add mutex for PulseAudio to avoid race conditions` and `audioio: protect PulseAudio global gconn with mutex against websocket restart race`) have since landed on `mercuryv2` independently, so this branch carries duplicates of them and they collide with what is already in the base.

Could you rebase onto current `mercuryv2` and drop those two commits? What should remain is just:

- `Add audio device matching by name`
- `audioio: fix name-matching deadlock and silent duplicate-name match`

I tried it locally and both cherry-pick cleanly onto current `mercuryv2` with no conflicts — the result is +136/-8 in `audioio/audioio.c` and a full `make` builds without warnings. So it should be a purely mechanical rebase.

If it is easier on your side, we are happy to do the rebase here instead — just say the word and we will take care of it.

One other thing: the earlier Windows x64 CI failure was not your code. The job died in `actions/checkout` with `Problem with the SSL CA cert` before it ever reached the compiler, so it should clear on a re-run.

ptsmonteiro:
If you guys can take it from here feel free. I'm kinda short on time.
Thank you!

rafael2k:
I'll take care of it. I'll implement and commit.

rafael2k:
Merged to `mercuryv2` directly — thanks Pedro, this is a genuinely useful quality-of-life fix. Your commit went in with your authorship intact (`cb3c2ff`), plus two follow-ups on top.

**What landed:** `facd375` on `mercuryv2` (`cb3c2ff` yours + `c33e877` + `facd375`).

Closing this PR rather than merging the branch because it needed a rebase onto 15 commits of `mercuryv2` (notably #151, which is what created the deadlock below) and two fixes.

### Two problems fixed on top

**1. Self-deadlock against #151.** `audioio_restart()` holds `s_pulse_lock` across the uninit → init cycle, and the new `resolve_device_string()` calls sit inside that region. They reach `get_soundcard_list()`, which takes the same non-recursive mutex, so any UI audio-config change under `-x pulse` hung mercury permanently with the audio threads already stopped:

```
ws_command_handler -> audioio_restart -> resolve_device_string
  -> get_soundcard_list -> pthread_mutex_lock(&s_pulse_lock)   [blocked]
```

Split into `get_soundcard_list_int(..., pulse_lock_held)` with the public function as a thin wrapper, so the one caller that already owns the lock skips it. Chosen over a recursive mutex to keep the lock structure explicit. This only bites on Linux/PulseAudio, which is why it wouldn't have shown up in your macOS testing.

**2. Exact-name matching stopped at the first hit** and reported `matches = 1`, so identical display names silently bound whichever device enumerated first. Not exotic — on my host *both* PulseAudio playback devices are named `Built-in Audio Analog Stereo`, and two of the same USB codec collide the same way. In a feature whose premise is that ids move around, silently picking one can key the wrong radio. It now counts all exact matches and warns, as your substring path already did.

I also managed to drop your no-match warning while restructuring that loop, and put it back in `facd375` — it's the most useful of the three messages, since otherwise a typo'd name surfaces only as `pa_stream_connect_playback: (3) Invalid argument`.

### Verified

On a host that happens to exercise every path:

```
-o 'Built-in Audio Analog Stereo'  -> WRN device name is ambiguous (2 matches) -- use an exact id from -z
-o 'mercurytestsink'               -> INF matched by name to 'MercuryTestSink' (id=mercurytest)
-o 'mfskA'                         -> exact id, passed through untouched
-o 'No Such Device'                -> WRN matches no known id or name ... run with -z to list valid devices
```

Deadlock reproduced first via a `futex_wait` backtrace on `s_pulse_lock`, then confirmed restart completes in ~10 ms and a later client is still served. Build clean, unit suite green.

One thing I did **not** carry over: the `mercury_engine.c` log line that echoes the requested capture/playback device. No objection to it, it just wasn't needed once the `audio-cap`/`audio-play` lines report `<id> '<name>'` directly — happy to take it if you'd rather have both.

Issue #150, opened 2026-08-03
pulseaudio/pipewire


pearcedotsgatgmaildotcom:
I am trying to get mercury/mercury-qt and hermes-gui going together.

However I am getting runtime errors (core dump)  with mercury doing this.

I have successfully managed to tx test frames using mercury -i pulse -o pulse -x pulse -G t
and receive them on a separate system thru a kiwisdr and pipewire/pulse using qtwgraph to do the audio routing using mercury -i pulse -o pulse -x pulse -G r .. 

I have mercury-qt running in a python virtual environment  and managed to connect using a browser .. but activating any of the menu options causes a mercury core dump 👎 

5:40:08.565 [+23.757s] [INF] [websocket] Client connected (conn 0x745578000ca0)
Segmentation fault (core dumped)

15:41:36.257 [+0.554s] [INF] [modem-rx] channel BUSY (floor=-74.4 dB)
0000   16 03 01 07 5f 01 00 07 5b 03 03 46 d0 21 42 11   ...._...[..F.!B.
0000   16 03 01 07 5f 01 00 07 5b 03 03 94 e5 16 73 9d   ...._...[.....s.
15:42:19.002 [+43.299s] [INF] [websocket] Client connected (conn 0x7ee1a0009270)
Segmentation fault (core dumped)
s@gt:~/Downloads/a/Rhizomatica/mercury$ 

15:43:14.813 [+0.541s] [INF] [modem-rx] channel BUSY (floor=-69.7 dB)
0000   16 03 01 07 5f 01 00 07 5b 03 03 1e 6d 0c 5d 2f   ...._...[...m.]/
0000   16 03 01 07 5f 01 00 07 5b 03 03 24 6d ec bb 10   ...._...[..$m...
15:43:47.405 [+33.134s] [INF] [websocket] Client connected (conn 0x79fcb8009270)
Segmentation fault (core dumped)

15:44:43.120 [+0.556s] [INF] [modem-rx] channel BUSY (floor=-59.3 dB)
0000   16 03 01 07 4d 01 00 07 49 03 03 8e bc cd f2 b0   ....M...I.......
0000   16 03 01 07 4d 01 00 07 49 03 03 65 e0 d6 b8 6d   ....M...I..e...m
15:44:54.427 [+11.862s] [INF] [websocket] Client connected (conn 0x71f898001030)
Segmentation fault (core dumped)

with mercury-qt

[...]

PR #151, opened 2026-08-03
audioio: add mutex for PulseAudio to avoid race conditions

pedromessetti:
Added s_pulse_lock mutex in audioio/audioio.c that serializes all PulseAudio global-state operations:
- get_soundcard_list(): holds the lock during the entire pulse_shared_init → device enumeration → cleanup sequence
- audioio_restart(): holds the lock during the pulse_shared_uninit → buffer clear → pulse_shared_init → thread creation sequence
- audioio_deinit(): holds the lock during pulse_shared_uninit

Also moved pulse_shared_uninit() out of audioio_stop_threads() and into audioio_restart() so the uninit+init pair stays under a single lock hold.

### DIFF
--- a/audioio/audioio.c
+++ b/audioio/audioio.c
@@ -125,6 +125,8 @@ static inline uint64_t audioio_monotonic_ms(void)
 }
 
 #if defined(__linux__)
+static pthread_mutex_t s_pulse_lock = PTHREAD_MUTEX_INITIALIZER;
+
 static bool pulse_init_already_initialized(const ffaudio_init_conf *aconf)
 {
     return aconf && aconf->error && strcmp(aconf->error, "already initialized") == 0;
@@ -1557,8 +1559,12 @@ int get_soundcard_list(int audio_system, int mode,
 #if defined(__linux__)
     if (audio_system == AUDIO_SUBSYSTEM_PULSE)
     {
+        pthread_mutex_lock(&s_pulse_lock);
         if (pulse_shared_init(&did_init) != 0)
+        {
+            pthread_mutex_unlock(&s_pulse_lock);
             return 0;
+        }
     }
     else
 #endif
@@ -1569,12 +1575,15 @@ int get_soundcard_list(int audio_system, int mode,
         did_init = true;
     }
 
-    // mode: FFAUDIO_DEV_PLAYBACK (0) or FFAUDIO_DEV_CAPTURE (1)
     ffaudio_dev *d = audio->dev_alloc(mode);
     if (d == NULL)
     {
         if (did_init)
             audio->uninit();
+#if defined(__linux__)
+        if (audio_system == AUDIO_SUBSYSTEM_PULSE)
+            pthread_mutex_unlock(&s_pulse_lock);
+#endif
         return 0;
     }
 
@@ -1603,6 +1612,10 @@ int get_soundcard_list(int audio_system, int mode,
     audio->dev_free(d);
     if (did_init)
         audio->uninit();
+#if defined(__linux__)
+    if (audio_system == AUDIO_SUBSYSTEM_PULSE)
+        pthread_mutex_unlock(&s_pulse_lock);
+#endif
     return count;
 }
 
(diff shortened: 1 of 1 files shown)


rafael2k:
Thanks Pedro — I built this and ran the suite (both green). The race you're closing is real, and I'd like to merge it. Three notes, one of which I think should be fixed before merge.

## It does not fix #150

Worth being explicit, because the two are easy to conflate. I tracked #150 down to a NULL dereference *inside* the libpulse callback: `pulse_dev_on_next_sink/source()` call `ffsz_dup(info->name)` and `ffsz_dup(info->description)` unchecked, and `ffsz_dup()` is `strlen()`-based, so a NULL field segfaults on libpulse's `threaded-ml` thread. Confirmed under gdb by forcing `description=NULL` at the callback:

```
#0  __strlen_avx2 ()
#1  ffsz_dup (sz=0x0) at ./ffaudio/ffbase/stringz.h:72
#2  pulse_dev_on_next_source (...) at ffaudio/ffaudio/pulse.c:289
```

A mutex around the callers can't prevent that — the crash happens with the lock held. Also, Stephen's crash needs no UI action at all, while the race here requires one.

The two changes are complementary, not competing: different files, and they merge with zero conflicts. Fix for #150 is in #153.

## The race is real and worth closing

`get_soundcard_list()` runs on the publisher thread the moment a UI client connects, while `audioio_restart()` runs on the websocket thread on `set_audio_config` and calls `pulse_shared_uninit()` — destroying the global `gconn` that enumeration is walking. Genuine use-after-free window. Good catch.

Moving `pulse_shared_uninit()` out of `audioio_stop_threads()` is safe: that function has exactly one caller (`audioio_restart`), and `audioio_deinit()` does its own uninit after joining.

## Please fix: asymmetric unlock leaks the mutex

The lock is taken if *either* subsystem is pulse (`audioio.c:1962`):

```c
if (audio_subsystem == AUDIO_SUBSYSTEM_PULSE || audio_subsys == AUDIO_SUBSYSTEM_PULSE)
    pthread_mutex_lock(&s_pulse_lock);
```

but the main-path unlock checks only the *new* one, and by then `audio_subsystem` has been reassigned (`audioio.c:2053`):

```c
if (audio_subsystem == AUDIO_SUBSYSTEM_PULSE)
    pthread_mutex_unlock(&s_pulse_lock);
```

So a **pulse → alsa** switch takes the lock and never releases it, and the next `get_soundcard_list()` blocks forever — the UI status feed dies permanently. The three early-exit branches (NULL/FIFO/SOCK) got this right with `was_pulse || ...`; only the main path didn't.

It's latent rather than live today — the sole caller always passes `ctx->audio_system`, which never changes — but it's a one-line fix: use the same `was_pulse || audio_subsystem == AUDIO_SUBSYSTEM_PULSE` condition as the other three exits.

## Minor: the commit message claims more than the patch does

It says it "serializes all PulseAudio global-state operations", but the capture and playback threads still call `audio->init()` → `ffpulse_init()` directly without the lock, and `ffpulse_init()`'s `if (gconn != NULL) return "already initialized"` is check-then-act. Those two threads are created back-to-back, so that's the most concurrent init pair in the program and it stays unserialized. Not a regression, and fine to leave for a follow-up — just worth not claiming it's covered.

With the unlock fix, this is good to merge.

rafael2k:
Verified 49276fa — that's the fix, thanks.

All four exit paths now use `was_pulse || audio_subsystem == AUDIO_SUBSYSTEM_PULSE`, which matches the lock condition `(old == PULSE || new == PULSE)` exactly, so the pulse → alsa path can no longer leak the mutex.

Checked here:

- build clean, unit suite green
- live test on `-x pulse -G`: triggered `set_audio_config` (so `audioio_restart()` runs through the locked uninit→init cycle), then connected a **second** UI client to force a fresh `get_soundcard_list()`. Both device lists came back immediately — if the mutex had leaked, the publisher thread would have blocked there permanently, so this exercises the exact failure mode.

LGTM, good to merge. The `ffpulse_init()` check-then-act between the capture and playback threads is still unserialized, but that's pre-existing and fine as a follow-up.

PR #155, opened 2026-08-05
arq/tnc: send REGISTERED <callsign> when both callsign set and listening

pedromessetti:
Send REGISTERED response on the control port when the client has set its callsign and enabled LISTEN ON.  Fires on three trigger points:

- MYCALL received after LISTEN ON: callsign set, listen already active
- LISTEN ON received after MYCALL: listen enabled, callsign already set
- TCP client reconnects: both conditions already satisfied

Adds send_registered callback through the arq_tnc notification seam.

### DIFF
--- a/data_interfaces/tcp_interfaces.c
+++ b/data_interfaces/tcp_interfaces.c
@@ -292,6 +292,10 @@ static void execute_control_command(char *buffer)
         memset(&cmd, 0, sizeof(cmd));
         cmd.type = ARQ_CMD_SET_CALLSIGN;
         snprintf(cmd.arg0, sizeof(cmd.arg0), "%s", tok);
+        /* Kept for the REGISTERED notification below: the secondary loop
+         * reuses cmd, and strtok_r advances past this token. */
+        char primary_call[sizeof(cmd.arg0)];
+        snprintf(primary_call, sizeof(primary_call), "%s", tok);
         if (arq_submit_tcp_cmd(&cmd) != 0)
         {
             tcp_write(CTL_TCP_PORT, (uint8_t *)"WRONG\r", 6);
@@ -308,6 +312,15 @@ static void execute_control_command(char *buffer)
         }
 
         tcp_write(CTL_TCP_PORT, (uint8_t *)"OK\r", 3);
+
+        /* REGISTERED is emitted here, on this thread, rather than from the ARQ
+         * event loop: OK goes straight out through tcp_write() while queued
+         * lines are flushed by the TNC writer thread, so a REGISTERED produced
+         * over there could reach the host BEFORE this OK -- the more secondary
+         * callsigns MYCALL carries, the wider that window gets.  Sending it
+         * after the write makes the order documented in docs/TNC.md hold by
+         * construction instead of by timing. */
+        tnc_send_registered(primary_call);
         return;
     }
 
--- a/data_interfaces/tcp_interfaces.c
+++ b/data_interfaces/tcp_interfaces.c
@@ -1336,6 +1336,14 @@ void tnc_send_disconnected()
         HLOGW("tcp-ctl", "Error queuing disconnected message");
 }
 
+void tnc_send_registered(const char *callsign)
+{
+    char buffer[128];
+    snprintf(buffer, sizeof(buffer), "REGISTERED %s\r", callsign);
+    if (tnc_queue_line_critical(buffer) < 0)
+        HLOGW("tcp-ctl", "Error queuing registered message");
+}
+
 void tnc_send_buffer(uint32_t bytes)
 {
     char buffer[64];
@@ -1358,6 +1366,7 @@ static const arq_tnc_callbacks_t g_arq_tnc_cbs = {
     .send_cqframe       = tnc_send_cqframe,
     .send_disconnected  = tnc_send_disconnected,
     .send_buffer        = tnc_send_buffer,
+    .send_registered    = tnc_send_registered,
 };
 
 void tnc_send_sn(float snr)
--- a/data_interfaces/tcp_interfaces.h
+++ b/data_interfaces/tcp_interfaces.h
@@ -53,6 +53,7 @@ void tnc_send_connected();
 void tnc_send_cqframe(const char *source_call, int bw_hz);
 void tnc_send_disconnected();
 void tnc_send_buffer(uint32_t bytes);
+void tnc_send_registered(const char *callsign);
 void tnc_send_sn(float snr);
 void tnc_send_busy(bool busy);
 void tnc_send_bitrate(uint32_t speed_level, uint32_t bps);
--- a/datalink_arq/arq.c
+++ b/datalink_arq/arq.c
@@ -414,22 +414,16 @@ static void handle_cmd(const arq_cmd_msg_t *msg)
     {
     case ARQ_CMD_SET_CALLSIGN:
     {
-        bool should_register = false;
         char call_copy[CALLSIGN_MAX_SIZE];
         pthread_mutex_lock(&g_conn_lock);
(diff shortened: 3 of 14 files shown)


rafael2k:
Reviewed against current `mercuryv2` (which now has the UI stack, so the real delta here is the two commits below). Built and ran the suite: clean, all green.

## `arq/tnc: send REGISTERED` — looks right, merge-ready

The locking is the part I'd normally expect to get wrong here, and it's correct: the callsign is copied into a local **under `g_conn_lock`**, and `arq_tnc_send_registered()` is called after the unlock. That matters because `tnc_send_registered()` goes through `tnc_queue_line_critical()`, which retries for up to 50 ms — holding `g_conn_lock` across that would stall every other command handler.

The 50 ms worst case on the ARQ event-loop thread is not a new risk either: `PENDING`, `CANCELPENDING`, `CONNECTED` and `DISCONNECTED` already block that same loop the same way, so this is consistent with what's there.

The three trigger points cover the orderings a host can produce, and the ordering guard (`MYCALL` only fires when `listen` is already set, `LISTEN ON` only when the callsign is already set) means the normal `MYCALL` → `LISTEN ON` sequence emits exactly one REGISTERED rather than two.

Two small things, neither blocking:

1. **Secondary callsigns don't trigger it.** `ARQ_CMD_ADD_SECONDARY_CALLSIGN` is the one callsign path that never emits REGISTERED. If a host adds a tactical address and expects the same acknowledgement per callsign, it won't get one. Deliberate?
2. **No test.** This is TNC wire behaviour, which is the kind of protocol detail we do normally pin — a check that REGISTERED appears once for each of the three orderings, and not before both conditions hold, would stop a future refactor from silently dropping it.

## `build: add SAN_LDFLAGS` — harmless, but I could not reproduce the failure

I tried to reproduce the undefined `__asan_register_globals` references on current trunk and could not:

- `make SANITIZE_ASAN_UBSAN=1 -j$(nproc)` from the top level (what `sanitizers.yml:94` runs): **exit 0**, zero undefined references. That path does build `utils` (`all: internal_deps utils`).
- `make -C utils SANITIZE_ASAN_UBSAN=1` on its own: also clean.

Looking at why: both `utils/Makefile` and `modem/freedv/Makefile` `include ../config.mk`, so with `SANITIZE_ASAN_UBSAN=1` passed on the command line (make forwards command-line variables to sub-makes) their `COMMON_CFLAGS` already carries `-fsanitize=address,undefined`. Every link line in both files passes `$(CFLAGS)`, and it is `-fsanitize` on the *link* line that pulls in the ASan runtime — so the flag is already there via `CFLAGS`, and `SAN_LDFLAGS` is currently redundant.

That said, it matches the convention documented in `config.mk` ("SAN_LDFLAGS, which every link line appends") and costs nothing, so I have no objection to merging it — I just can't confirm it fixes a live breakage.

Could you link the failing CI run? If it fails in a configuration I'm not hitting (different compiler, or a target `all` doesn't reach such as the `freedv_data_*` binaries), that's worth knowing, because then the same gap probably exists in other sub-makes too and we'd want to fix them all rather than these two.

**Verified here:** build clean, unit suite green on the PR head.

rafael2k:
Re-reviewed `eec0c52` + `775883d`. The decoupling is exactly right, and the extras beyond what I asked for — `docs/TNC.md` in all four places, the unit test, and moving `drainAfter` in the integration tests — are appreciated. Build clean and the full unit suite green here.

One issue left, and it is one the new documentation introduces.

## The documented OK → REGISTERED ordering is not guaranteed

`docs/TNC.md` now states *"Arrives after the `OK\r` response to `MYCALL`"*, and the session-flow diagram shows `OK` then `REGISTERED`. The code cannot promise that, because the two lines travel different paths on different threads:

- `OK` is a **direct socket write**: `tcp_write(CTL_TCP_PORT, "OK\r", 3)` → `net.c:235`
- `REGISTERED` is **queued**: ARQ event loop → `arq_tnc_send_registered()` → `tnc_queue_line_critical()` → `tnc_tx_chan`, drained by the TNC writer thread

And in the `MYCALL` handler the command is submitted *before* the OK is written:

```c
cmd.type = ARQ_CMD_SET_CALLSIGN;
if (arq_submit_tcp_cmd(&cmd) != 0) { ... }      /* ARQ loop can run REGISTERED from here */

while ((tok = strtok_r(NULL, ...)) != NULL)      /* secondary callsigns widen the window */
    arq_submit_tcp_cmd(&cmd);

tcp_write(CTL_TCP_PORT, (uint8_t *)"OK\r", 3);   /* ...only now does OK go out */
```

So between the submit and the `tcp_write`, the ARQ loop can dispatch `SET_CALLSIGN` and the writer thread can flush `REGISTERED` first. The window is small but real, and `MYCALL PRIMARY SEC1 SEC2 ...` widens it — each secondary token is another submit before the OK. A host reading line-by-line could then see `REGISTERED <call>` as the first line after `MYCALL`, which is precisely what a VARA host does not expect.

**Suggested fix:** emit REGISTERED from the `MYCALL` handler in `tcp_interfaces.c`, immediately after the OK write. It is a pure function of the callsign just accepted, so it does not need to come from the ARQ thread, and this makes the documented order true by construction rather than by timing. The `CLIENT_CONNECT` trigger can stay exactly where it is — nothing is racing it there.

(Moving the `tcp_write` up to just after the primary submit would narrow the window but not close it, so I would not do that instead.)

## Everything else

- `SET_CALLSIGN`: the callsign is still copied under `g_conn_lock` and sent after the unlock — good.
- `CLIENT_CONNECT`: gated on callsign only now, correct. A reconnecting host that then re-sends `MYCALL` gets two REGISTEREDs; harmless.
- `test_tnc_send_registered` pins the exact wire format including the `\r` — that is the right level for this.
- `ARQ_CMD_ADD_SECONDARY_CALLSIGN` still never emits REGISTERED. Fine if VARA only registers the primary; worth a one-line comment saying so, otherwise the next reader will wonder.

With the ordering fix this is good to merge. The `SAN_LDFLAGS` commit is unchanged and my earlier question still stands — a link to the failing CI run whenever you have it.

rafael2k:
Pushed the ordering fix onto this branch as `512d13e` so you don't have to round-trip on it — please sanity-check it.

**What changed:** REGISTERED now goes out from the `MYCALL` handler in `tcp_interfaces.c`, immediately after the `tcp_write("OK\r")`, instead of from the ARQ event loop. It's a pure function of the callsign just accepted, so it doesn't need the ARQ thread, and the order your `docs/TNC.md` documents now holds by construction instead of by timing. Removed the emit from `ARQ_CMD_SET_CALLSIGN`; the reconnect trigger stays exactly where you put it, since nothing races it there. `MYCALL` is the only submitter of `ARQ_CMD_SET_CALLSIGN`, so no path loses its notification.

**The test pins the order, not just the content.** The chan stub now records how many `tcp_write()` calls had already happened when a line was queued, and the case uses `MYCALL TEST1 SEC1 SEC2` — the secondary tokens are what widen the window the race needed.

I checked it actually catches the regression rather than passing vacuously: with your original code it fails —

```
test_cmd_mycall_registered_follows_ok:FAIL: Expected 'REGISTERED TEST1\r' Was ''
```

— and passes with the fix. Worth knowing that the first time I ran that negative check it *passed*, because `tests/` does not rebuild when a source outside that directory changes. A `make clean` in `tests/` is needed, otherwise you can get a green run against a stale binary. Might be worth fixing in the tests Makefile separately.

Full build and unit suite green from clean. Your `docs/TNC.md` wording needed no change — it was already correct, the code just had to catch up with it.

Everything else in the PR looks good to me. Still curious about the failing CI run for the `SAN_LDFLAGS` commit whenever you have the link.

PR #158, opened 2026-08-05
arq/tnc: acknowledge every callsign MYCALL accepts, not just the primary

rafael2k:
Follow-up to #155, from VARA's own documentation:

```
REGISTERED Call<cr>     Call sign registered in VARA
```

It is defined **per call sign**. `MYCALL` can carry secondaries and Mercury answers for all of them, so a host that registered three callsigns and got a single `REGISTERED` back had no reason to believe the other two were live.

Now one line per accepted callsign, primary first, in the order given:

```
MYCALL AAAA BBBB CCCC\r
OK\r
REGISTERED AAAA\r
REGISTERED BBBB\r
REGISTERED CCCC\r
```

Only the ones the ARQ layer actually took. Past `CALLSIGN_MAX_SECONDARY` it drops them, and acknowledging a callsign we will never answer is worse than staying quiet — the host would address a station that ignores it.

**Tests:** three callsigns produce three lines in order; six secondaries produce `1 + CALLSIGN_MAX_SECONDARY`. The harness now records the whole queued sequence instead of only the last line, which is also what the existing ordering test really meant to assert (it was checking `last_queued_line` for the primary, which only held while the primary was the sole line).

`docs/TNC.md` updated with the multi-callsign form and the drop rule.

**Not implemented here:** VARA's `LINK REGISTERED` / `LINK UNREGISTERED`, which report the *remote* station's status during a link — a separate feature, happy to add it if wanted.

Build clean, unit suite green.

### DIFF
--- a/data_interfaces/tcp_interfaces.c
+++ b/data_interfaces/tcp_interfaces.c
@@ -302,13 +302,21 @@ static void execute_control_command(char *buffer)
             return;
         }
 
-        /* Secondary callsigns (remaining tokens) */
+        /* Secondary callsigns (remaining tokens).  Accepted ones are kept so
+         * each can be acknowledged below: "REGISTERED <Call>" is per call
+         * sign in VARA, and Mercury answers for every one of these. */
+        char accepted_secondaries[CALLSIGN_MAX_SECONDARY][CALLSIGN_MAX_SIZE];
+        int  accepted_count = 0;
         while ((tok = strtok_r(NULL, " \t\r\n", &saveptr)) != NULL)
         {
             memset(&cmd, 0, sizeof(cmd));
             cmd.type = ARQ_CMD_ADD_SECONDARY_CALLSIGN;
             snprintf(cmd.arg0, sizeof(cmd.arg0), "%s", tok);
-            arq_submit_tcp_cmd(&cmd);   /* best-effort; overflow is logged in arq.c */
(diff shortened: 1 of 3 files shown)


rafael2k:
Re-reviewed against current trunk (`f79a28e`). CI is green (8/8). **Merge-ready.**

I checked the one thing that could have made this wrong — whether the cap at `CALLSIGN_MAX_SECONDARY` actually matches what the ARQ layer keeps:

- `arq_submit_tcp_cmd()` returns 0 only when the command was accepted onto the bus, so the `!= 0 → continue` guard is the right polarity and a rejected submit correctly goes unacknowledged.
- `ARQ_CMD_ADD_SECONDARY_CALLSIGN` keeps the **first** `CALLSIGN_MAX_SECONDARY` (4) and logs the rest as dropped, so acknowledging exactly `min(count, 4)` mirrors the real behaviour rather than merely approximating it.
- `ARQ_CMD_SET_CALLSIGN` clears `secondary_call_count` back to 0, and the bus is FIFO with a single consumer, so the `MYCALL` → `ADD_SECONDARY×N` sequence always starts from an empty list. A second `MYCALL` cannot silently overflow and leave us claiming registrations the ARQ layer discarded — which was the failure mode worth checking.

Neither side de-duplicates, so `MYCALL A B B` acknowledges `B` twice and stores it twice. Consistent between the two, so not a defect introduced here.

PR #159, opened 2026-08-05
Add configurable tx_delay_ms

### LINKED ISSUE #154: PTT vs tx audio delay
I've been trying to test this modem between different radios (high snr) and the connection handshake works fine. However, the data frames are almost never decoded at reception. I observe ISS repeating the transmission and no decode on IRS. 

In my particular case the PTT is activated after receiving PTT ON from the API instead of having mercury configured to trigger the PTT. But this setup should work because it works if I replace mercury with vara. I believe we should have some tx_audio_delay config setting to account for the transceiver tx mode to be fully active as I'm suspecting that the problem is that the transmitted frames are clipped in the beginning.

ptsmonteiro:
Adds a configuration setting to set the delay between PTT ON and actual audio being sent. 

Fixes #154.

### DIFF
--- a/common/cfg_utils.c
+++ b/common/cfg_utils.c
@@ -54,6 +54,7 @@ void cfg_set_defaults(mercury_config *cfg)
     cfg->tnc_keepalive_s = 60;
     cfg->tnc_buffer_report_ms = 1000;
     cfg->tx_gain_db            = 0.0f;
+    cfg->tx_delay_ms           = 10;
     cfg->data_retry_slots            = ARQ_DATA_RETRY_SLOTS_DEFAULT;
     cfg->mode_hold_after_downgrade_s = ARQ_MODE_HOLD_AFTER_DOWNGRADE_S_DEFAULT;
     cfg->ladder_up_successes         = ARQ_LADDER_UP_SUCCESSES_DEFAULT;
@@ -243,6 +244,9 @@ bool cfg_read(mercury_config *cfg, const char *ini_path)
     if (d >  20.0) d =  20.0;
     cfg->tx_gain_db = (float)d;
 
+    i = iniparser_getint(ini, CFG_KEY_TX_DELAY_MS, cfg->tx_delay_ms);
+    if (i >= 0 && i <= 2000) cfg->tx_delay_ms = i;
+
     iniparser_freedict(ini);
     return true;
 }
@@ -348,6 +352,7 @@ bool cfg_write(const mercury_config *cfg, const char *ini_path)
 
     fprintf(f, "\n[audio]\n");
     fprintf(f, "tx_gain_db = %.2f\n", cfg->tx_gain_db);
+    fprintf(f, "tx_delay_ms = %d\n", cfg->tx_delay_ms);
 
     fprintf(f, "\n[tnc]\n");
     fprintf(f, "keepalive_s = %d\n", cfg->tnc_keepalive_s);
--- a/common/cfg_utils.h
+++ b/common/cfg_utils.h
@@ -44,6 +44,7 @@
 #define CFG_KEY_NO_PROGRESS_TIMEOUT_S "arq:no_progress_timeout_s"
 #define CFG_KEY_DISCONNECT_DRAIN_TIMEOUT_S "arq:disconnect_drain_timeout_s"
 #define CFG_KEY_TX_GAIN_DB          "audio:tx_gain_db"
+#define CFG_KEY_TX_DELAY_MS         "audio:tx_delay_ms"
 #define CFG_KEY_TNC_KEEPALIVE_S     "tnc:keepalive_s"
 #define CFG_KEY_TNC_BUFFER_REPORT_MS "tnc:buffer_report_ms"
 #define CFG_KEY_ARQ_DATA_RETRY_SLOTS          "arq:data_retry_slots"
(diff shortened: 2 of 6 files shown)


rafael2k:
Merges cleanly onto current `mercuryv2` (17 commits behind, no conflicts), builds clean, unit suite green. The shape is right and I'd like to take it — one substantive issue to settle first.

### The delay is invisible to the ARQ turnaround budget

`tx_delay_ms` is clamped to 0..2000 and inserted after `ptt_on()` / `arq_modem_ptt_on()`, before the audio goes to the ring. Nothing in `datalink_arq/` knows it exists — `grep -rn tx_delay datalink_arq/` is empty. Meanwhile the guards it competes with are:

```
channel_guard_ms      = 700
iss_post_ack_guard_ms = 900
```

So a `tx_delay_ms` anywhere near those values silently eats the turnaround budget at **both** ends: every burst starts later than the peer's listen window assumes, and the failure mode is not "PTT is late", it is ACK timeouts and retransmissions that look like a bad channel. At the 2000 ms ceiling the mode is simply broken, with no diagnostic pointing back at this setting.

Three ways out, in my order of preference:

1. **Fold it into the guard** — have `arq_protocol` add `modem_get_tx_delay_ms()` to `ARQ_CHANNEL_GUARD_MS`, so the timing budget stays correct at any value. Most correct, and it makes the knob genuinely safe.
2. **Clamp much lower** (say 0..500) and say why in the header comment.
3. At minimum, **warn at startup** when `tx_delay_ms` exceeds the channel guard, so the eventual "why is ARQ retrying" has a breadcrumb.

I lean to (1); happy to do that on top if you'd rather not.

### Placement is correct

Worth recording, since it is the subtle part and it is right: `ptt_on()` and `arq_modem_ptt_on()` both fire *before* the sleep, so the TNC "PTT ON" notification is already out when the delay starts. That is exactly what the slow-external-PTT use case needs. And the delay is only in the real-audio `else` branch — the virtual-clock/sock path is untouched, so the deterministic sim and integration harnesses are unaffected. Good instinct.

### Two minor things

- **Out-of-range ini values are silently ignored** rather than clamped:
  ```c
  i = iniparser_getint(ini, CFG_KEY_TX_DELAY_MS, cfg->tx_delay_ms);
  if (i >= 0 && i <= 2000) cfg->tx_delay_ms = i;
  ```
  `tx_gain_db` immediately above clamps instead. Someone writing `tx_delay_ms = 5000` gets the 10 ms default and no explanation. `modem_set_tx_delay_ms()` clamps anyway, so clamping here too would be consistent and predictable.

- `[audio]` is a slightly odd home for what is really a **keying**/PTT parameter — `[radio]` would read better next to the other radio-control settings. Cosmetic, and not worth a wire of churn if you disagree.

Default stays 10 ms, so no behaviour change for anyone who does not set it. Once the guard interaction is resolved this is a merge from me.

rafael2k:
Follow-up after checking what already exists, since this is not adding a new concept — it is parameterising one we already have.

**The delay already exists, hardcoded.** `modem.c` has had `usleep(10000)` after `ptt_on()` in the data TX path all along, with the comment "Wait for radio relay to switch (10ms for your radio)". So this PR is the right change in principle: that comment literally says "your radio", which is exactly the thing that should not be baked into a constant.

**But the tune carrier has no settle delay at all.** The tune thread does:

```c
ptt_on();

while (!shutdown_ && !atomic_load(&g_tune_stop))
{
    ...
    write_buffer(playback_buffer, (uint8_t *)chunk, sizeof(chunk));
```

No wait between keying and the first sample. Today that is a cosmetic inconsistency (10 ms on data, 0 on tune). With this PR it becomes a functional gap: someone who sets `tx_delay_ms = 500` for a slow external PTT gets a correctly-delayed data path, then presses TUNE and transmits into a radio that has not keyed yet — and feeding the ATU is the entire purpose of that carrier. Whatever value we settle on should apply in both places.

**Hamlib already has a per-rig version of this and we never set it.** `rig.h` carries `post_ptt_delay` — "delay after PTT to allow for relays and such" — and `grep -rn post_ptt_delay radio_io/*.c` comes back empty. For hamlib-controlled rigs the backend already knows the rig's relay timing, so there is an argument that the correct fix for that class of radio is to honour hamlib rather than ask the operator. It does not replace this PR — the stated use case is a *host app* driving PTT off the TNC notification, which hamlib knows nothing about — but it is worth deciding deliberately rather than ending up with two mechanisms that disagree.

None of this changes my earlier point about the ARQ guard budget, which is still the blocking one.

rafael2k:
Merged — thanks Pedro, nice piece of work.

The thing that made this easy to say yes to: the default is *exactly* today's behaviour. `cfg->tx_delay_ms = 10`, `g_tx_delay_ms = 10`, and the data path becomes `usleep(modem_get_tx_delay_ms() * 1000)` — byte-for-byte the `usleep(10000)` that was already there. Nobody who does not set the key notices anything, which is what makes the ARQ-guard question I raised a matter for the operator who deliberately raises it rather than a blocker.

You also put the sleep in the right place, which is the subtle part: `ptt_on()` and `arq_modem_ptt_on()` both fire *before* it, so the TNC "PTT ON" notification is already out when the delay starts — exactly what a host app driving an external relay needs. And it only touches the real-audio branch, leaving the virtual-clock/sock path alone, so the deterministic sim and integration harnesses are untouched. That was the correct instinct.

Three small follow-ups I'll pick up rather than hold the merge for:

1. **Startup warning when `tx_delay_ms` exceeds the channel guard (700 ms)** — not to stop anyone, but because the failure mode is ACK timeouts that look exactly like a bad channel, with nothing pointing back at the setting.
2. **`mercury.ini.example` says "values outside are clamped"**, but `cfg_read` does `if (i >= 0 && i <= 2000) cfg->tx_delay_ms = i;` — out-of-range values are *ignored*, so `tx_delay_ms = 5000` silently gives 10 where the docs promise 2000. Will clamp, matching `tx_gain_db` right above it.
3. **The tune carrier has no settle delay at all** — it calls `ptt_on()` and immediately writes the first chunk. Harmless today, but with your knob a slow external PTT would transmit the ATU carrier into an unkeyed radio. The delay should apply there too.

Separately: hamlib has a per-rig `post_ptt_delay` ("delay after PTT to allow for relays and such") that we have never read. I have added log-only reporting of it in `a08216a` to find out whether real radios report anything useful. The dummy backend says 0. If real rigs also say 0, your configurable knob is straightforwardly the right mechanism and there is no competing one to reconcile — which is the outcome I expect.

Worth noting for context: our 10 ms and the ARQ guards were tuned for the sbitx, which keys over the SHM interface and never goes near hamlib at all.

PR #160, opened 2026-08-05
arq+mfsk: make the MFSK robust rung actually carry a transfer

rafael2k:
Supersedes #125. Goal: the MFSK weak-signal rung as the ladder floor, with a data plane that works on a real-time audio path.

> **Status (2026-08-26): both blockers that held this PR are fixed.** The branch now carries a 1 KB transfer, and the fading failure has been root-caused rather than retuned. It beats trunk below ~5 dB SNR and is slower above it, by a fixed and now well-characterised margin. Details below; the per-change working is in the comments.

## Where it stands

Symmetric ITU-R `good` fading, Go integration harness driven by `MERCURY_BIN`, wall-clock including connect, both builds on the same trunk base:

| payload | SNR3k | trunk | this branch | |
|---|---|---|---|---|
| 1 KB | 10.2 dB | 82 s | 138 s | trunk 1.7× |
| 1 KB | 5.2 dB | 100 s | 138 s | trunk 1.4× |
| 102 B | 0.2 dB | 116 s | **100 s** | branch 1.2× |
| 102 B | −2.8 dB | 145 s | **100 s** | branch 1.5× |
| 102 B | −4.8 dB | 226 s | **99 s** | **branch 2.3×** |

The branch is essentially **flat at 99–138 s across the whole range** while trunk degrades from 82 s to 226 s. Crossover between 5.2 and 0.2 dB. That shape is the point of the work: a floor rung that keeps working when the channel does not.

## The two blockers, closed

**The fading failure was never a fading regression.** The branch could not connect over a faded reverse path where trunk could — 0/4 at 10.2 dB against trunk's 2/2. Bisect showed the trunk this branch forked from (`bf7d546`) fails the same test 0/4, so it was never introduced here; trunk fixed it in `c70fdd6` (anchor the CALL retry to PTT-OFF) and this branch already contained that fix. The same class of defect also existed on the *answerer* side and only became reachable when the ACCEPT window doubled to cover a 13.5 s MFSK burst:

```
+3.617   B decodes CALL#1
+4.349   B sends ACCEPT#1          ← one channel guard later, correct
+15.361  B decodes CALL#2, logs it, AND DOES NOTHING
+26.161  B re-sends ACCEPT#2       ← 18.1 s later, the data-burst window
+23.617..+27.327   A is transmitting CALL#3
```

The retransmitted ACCEPT landed inside the caller's own burst every time. A CALL arriving is proof the caller has *not* started a data burst, so there is nothing for the long window to protect. Fixed in `21c01f3` by re-anchoring to one channel guard from the CALL just heard. Connect goes 0/4 → 2/2, parity with trunk from 15.2 down to 0.2 dB.

**The 1 KB stall is gone.** 1054/1054 bytes now deliver at every SNR where the link closes at all.

## Ladder tuning

Two follow-ups, each measured on a good-link *and* a fringe cell — the good-link cell alone will happily claim an improvement:

- **`fdf44f7`** — downgrade threshold 2 → 4, sized from the payload ratio between neighbouring rungs rather than from how bad a retry feels. DATAC1 carries 4× DATAC3, so it keeps winning until it misses ~3 bursts in 4. **234 s → 148 s**, fringe unchanged.
- **`61179cf`** — DATAC15 dropped from the ladder. It carried 6.82 B/s against the MFSK floor's 7.26 *and* was the less robust of the two: worse on both axes the ladder is ordered by, so never the best choice at any SNR, yet every climb paid a burst and a turnaround to pass through it. **148 s → 137 s**, fringe unchanged. `test_ladder_table_ordered_and_sized` now asserts the ordering property, not just the contents.

## The one remaining gap

The branch costs the same 138 s at 10.2 dB and at 5.2 dB — it does not get faster when the channel gets better. That is a fixed startup tax: the blind climb from the MFSK floor, three bursts carrying 278 B between them before DATAC1 is reached, whatever the channel would have supported immediately.

Four shortcuts were tried and **all four were measured on both cells and rejected**: a two-rung fast ramp (at either threshold: 300 s timeout), an SNR-derived climb ceiling (372 s), and raising the start rung with either ladder (159 s good / **185 s fringe against 99 s**). Bigger jumps skip DATAC1, the rung that actually pays, and amplify ISS/IRS mirror mismatch; the ceiling ruled out the productive bursts; a raised start rung wrecks the fringe, exactly as `ARQ_LADDER_START_LEVEL`'s comment warns.

So the tax is structural, not a constant to tune. The ladder must *discover* the channel by climbing because it deliberately has no SNR input, where trunk skips discovery using its estimate. Closing it means letting both ends **agree** a start rung — feasible without giving up the delivery-driven design, since the control plane always decodes DATAC16 and the frame header already carries an SNR field, so the ISS can announce its opening rung before the first DATA burst. It cannot be done unilaterally: the IRS binds its payload decoder to the mirrored rung, so a higher first burst would go out in a mode the receiver is not listening for. That is a new control message, and it is the one change that would put this branch ahead of trunk everywhere rather than below ~5 dB.

## The engineering

Seven bugs found and fixed, each with a test that fails without the fix:

| commit | bug |
|---|---|
| `2058c43` | `ARQ_ACCEPT_RX_WINDOW_MS` fixed at 9000 ms, sized for a DATAC15 first frame. The ladder floor is MFSK at 13.5 s, so the IRS gave up mid-burst and retransmitted ACCEPT **on top of** the ISS's data, going deaf for 3.7 s of every burst. Now derived from the ladder. |
| `c33e353` | `mfsk_downmix()` malloc'd and rebuilt the whole window per search — 1.7 MB, 215k trig calls, a 63-tap FIR over 107k samples, ~6×/s. Now incremental, carrier phase from the absolute sample index. |
| `eb31722` | **The modulator clipped 45% of its own payload**: `MFSK_TXAMP` 6000 against a 32767 rail gave a true peak of 42426 and PAPR 1.9 dB. Round-trip tests could not see it — both ends saw the same distortion. |
| `eb31722` | RX window sized for two bursts; with one, a burst was decodable for only ~0.48 s of slide. |
| `b1c527e` | Sync search trimmed to offsets that can actually decode; postamble fallback gated. |
| `c386102` | Backlog cap and ACCEPT window both derived from `arq_protocol_longest_burst_s()`. The cap was recomputed from the **live** payload mode and shrank to ~7 s mid-session, flushing bursts still arriving. |
| `eae2459` | `expect_pattern_ack` was true for all of CONNECTED, and modem.c uses it to gate a ~3-burst correlation run on **every capture chunk** — a detector for a signal that can only arrive in one state, running all session, eating half the receive budget. |
| `21c01f3` | The answerer ignored a repeated CALL and waited out the 18 s data-burst window, so every retransmitted ACCEPT collided with the caller's next CALL. |

Test additions: `tests/modem/test_mfsk_modem.c` feeds the way the live funnel does (880-sample chunks carrying the remainder — 880 is not a multiple of nin=320, and a loop ignoring that silently drops 240 samples per chunk), sweeps prior audio 0.08–25.6 s and trailing audio to 16 s, and asserts anti-clipping on the TX samples directly, which is the only way to catch a defect a round trip cannot see. `tests/datalink_arq/test_arq_ladder.c` pins the ladder-derived guards against every rung and now the goodput ordering itself.

Full unit suite and the Go integration harness green.

## Still open

- **32-bit ARM unmeasured.** Headroom on a laptop is 6.4–8.0k samp/s against 8k needed. If it is tight on a Pi, the lever is per-tone Goertzel detection (see `modem73/phy/mfsk_modem.hh`) rather than downmix + FIR + correlation.
- We key hard on/off; a raised-cosine edge ramp is worth taking regardless.
- On-air validation. Everything above is one channel model on one host; absolute dB from watterson are not trustworthy, and the comparisons are relative at identical settings.

## Superseded measurements

The earlier version of this description reported the branch losing under fading (1/5 against trunk's 5/5) and stalling at 276 bytes on a 1 KB transfer. Both were real at the time and both are fixed; the fading one was a defect in the connect handshake, not in the fading performance of the mode. Its AWGN and asymmetric-link sweeps predate the ladder changes above and should be re-run before being quoted.


rafael2k:
## The fringe floor was the sync accept threshold, not the code

Pushed `453d509`. Following up on the sensitivity question — the answer turned out not to be the LDPC, the LLRs, or anything in the demodulator.

`mfsk_sync_search` ended with a magic constant:

```c
return (best_fine_metric < 0.5) ? -1 : best_fine;
```

The per-symbol statistic is `|corr|²/(E_tmpl·E_rx) = SNR_sym/(1+SNR_sym)`, so a 0.5 gate demands **SNR_sym ≥ 0 dB**. That one number was the weak-signal limit of the entire mode — everything below it was thrown away before the demodulator ran.

### The proof: the code doesn't matter

Sweeping all five ported LDPC rates gave **byte-identical** FER curves (2/20 and 17/20 at the same SNRs):

| rate | K | payload | FER @ −1.1 dB | FER @ −0.1 dB |
|---|---|---|---|---|
| 8/16 | 800 | 100 B | 0.90 | 0.15 |
| 5/16 | 500 | 62 B | 0.90 | 0.15 |
| 3/16 | 300 | 37 B | 0.90 | 0.15 |
| 2/16 | 200 | 25 B | 0.90 | 0.15 |
| 1/16 | 100 | 12 B | 0.90 | 0.15 |

A 16× stronger code changing *nothing* means the limiter is upstream. Instrumenting the search confirmed it: at SNR3k −3.1 dB it returns `off=-1`, `metric=0.344`, payload never attempted.

### What the detector actually sees

| input | metric |
|---|---|
| **noise only** (600 searches, flat vs SNR) | 0.037 – 0.049 |
| signal @ −5.1 dB | 0.219 |
| signal @ −9.1 dB | 0.091 |

The signal stood ~10× above the noise floor while the gate sat *above the signal*. `0.08` is 1.6× the measured noise maximum — **0 false syncs in 600 noise-only searches**.

### Results

| | before | after |
|---|---|---|
| AWGN 50% FER | −0.6 dB | **−11.2 dB** (~10.6 dB) |
| Watterson 2 ms/1 Hz 50% FER | +5.0 dB | **−3.7 dB** (~8.7 dB) |
| Watterson error floor | 0.25–0.30, never cleared | **gone** (0.00 at +3.9 dB) |

That floor was sync failing during fades, not the decoder. **DATAC15 measures −10.3 dB on the same harness**, so the ladder floor now sits ~0.9 dB *below* the rung above it — which is the whole point of having a fringe rung.

### The hazard that had to be fixed with it

A lower threshold admits more bad anchors, and they were not survivable: `modem_mfsk` caches the located anchor and marks it tried, so a false peak whose payload failed CRC left the decoder **blind until it slid out of the window** — a missed burst, not just wasted CPU. A resident payload that fails CRC now releases the anchor and the search restarts one symbol past it (safe: the search returns the *earliest* peak above threshold, so nothing decodable can hide before it).

### Tests

Both new tests were verified to fail without their fix — the weak-preamble test fails when rebuilt with `-DMFSK_SYNC_ACCEPT=0.5`, and the failed-anchor test fails when the anchor release is removed. Neither burst in the latter carries a postamble; with one, the postamble fallback masks the defect completely (my first version of that test passed against unfixed code for exactly this reason).

Full C suite green, `make integration-test` green (bidirectional ARQ 112/112 both directions).

### What I measured and deliberately did *not* ship

Rewriting the LLR derivation to codec2's non-coherent model (log-I₀ of amplitude + `max*` combining, replacing an energy difference with a ±5 clamp) moved the cliff by **~0 dB** at both the old and the new threshold — byte-identical FER over 30 trials/point — while costing ~6% more decode time, which this mode cannot spare on 32-bit ARM.

### Where the next dB are

Below ~−12 dB the signal metric reaches the detector's own noise floor, so no threshold goes further. That needs a detector that **combines** the preamble symbols instead of averaging per-symbol normalised ratios — averaging gains nothing from preamble length. Rate 1/8 buys only ~0.5 dB there, so the rung stays at rate 1/2 (all five codes share N=1600: rate costs payload, not airtime).

rafael2k:
## Second pass: the detector is not the bottleneck — measured, so I stopped there

Following up on "improve the detector too". I built the detector improvement, measured it, and it does not earn its place. Reverted; `453d509` remains the only shipped change. Here is the evidence, because the negative results are the useful part.

### The oracle settles it

Bypassing the search entirely and decoding at the **known true preamble offset**:

| | search-based | oracle (perfect timing) |
|---|---|---|
| rate 1/16 @ −13.1 dB | 15/20 | 16/20 |
| rate 1/16 @ −14.1 dB | 1/20 | 2/20 |

Perfect knowledge of the burst position buys essentially nothing. Removing the accept gate altogether (CRC as the only detector) buys ~0.5 dB at −11 dB and does **not** move the −12 dB wall at all. Below ~−12 dB the limiter is the demod/decode chain, not acquisition.

I did write a ranked top-K candidate API (`mfsk_sync_search_candidates`) plus speculative decoding, since at the fringe the true peak still *ranks* first even when it no longer stands above the noise maximum. The oracle shows there is nothing to win, so it went in the bin rather than into the modem — it would have cost LDPC decodes on noise while idle, and MFSK RX CPU is already the live-path constraint.

### The design curve — tone count beats code rate

AWGN, 50% FER, gate removed:

| config | burst | goodput | 50% FER |
|---|---|---|---|
| **M=32, rate 1/2** (current) | 13.1 s | 59.8 bps | −11.4 dB |
| M=16, rate 1/2 | 16.3 s | 48.0 bps | −12.2 dB |
| M=8, rate 1/2 | 22.0 s | 35.6 bps | −12.9 dB |
| M=4, rate 1/2 | 32.6 s | 24.0 bps | −13.5 dB |
| M=8, rate 3/16 | 22.0 s | 12.7 bps | **−14.1 dB** |
| M=32, rate 1/16 | 13.1 s | 6.1 bps | −13.5 dB |

Two things fall out. **9 dB of code-rate reduction (1/2 → 1/16) buys only 2.1 dB** — nowhere near theory. And M=4/rate-1/2 reaches the *same* −13.5 dB as M=32/rate-1/16 with **4× the goodput**, so the code rate is strictly the worse lever. M=8/rate-3/16 is the first config that actually reaches the −14 dB figure.

LLR quality is not the floor either: log-I₀ + `max*`, and widening the ±5 clamp to ±40, both change **nothing** even at rate 1/16.

### …but fading inverts the ranking, so I am not proposing the change

Watterson 2 ms/1 Hz, rate 1/2, shipping gate:

| | 50% FER | error floor |
|---|---|---|
| **M=32** | −4.5 dB | **0.00** |
| M=8 | −5.8 dB | 0.10 |
| M=4 | −7.3 dB | 0.30–0.40, non-monotonic |

A 22–33 s burst spans many 1 Hz fade cycles, so a deep fade kills the frame at *any* SNR — M=4 sits at 30–40% FER even on a good channel. M=8 would trade a 10% error floor and 40% of the goodput for ~1.3 dB. That is a bad deal for an ARQ rung, so **M=32 stays**.

### Where the next real gain is

There is no time-interleaving across the burst. That is precisely why the lower-rate and lower-M configurations cannot cash in their coding gain on a fading channel — the code sees a contiguous wipe-out rather than spread errors. Interleaving is the change that would make the deeper configs viable; the detector and the code rate are both exhausted.

Worth noting the shipped fix already removed the *sync-caused* fading error floor (0.25–0.30 → 0.00 at M=32). What remains is genuine fade loss.

rafael2k:
## Time-interleaving implemented — `4579a69`

An LDPC block code corrects scattered errors far better than a contiguous run, and a fading HF channel delivers exactly the contiguous case: the burst is 13.5 s, so a deep fade wipes a stretch of consecutive symbols and, with the coded bits laid down in order, a stretch of consecutive codeword bits with it. The decoder was being handed the one error pattern it is worst at.

Transmitted slot `t` now carries coded bit `ilv[t]` — Fisher-Yates from a fixed xorshift seed, integer arithmetic only, so both ends derive an identical table on any platform with nothing sent over the air. Pseudo-random rather than a fixed stride: these are quasi-cyclic codes (N=1600, circulant 100) and an arithmetic progression risks landing the damage on the same check nodes.

### Measured — 60 trials/point, M=32 rate 1/2, builds verified to differ only in the interleaver

| channel | SNR3k | without | **with** |
|---|---|---|---|
| Watterson 2 ms / 1 Hz | −3.1 dB | 0.35 | **0.20** |
| Watterson 2 ms / 1 Hz | −0.1 dB | 0.08 | 0.07 |
| Watterson 2 ms / 0.2 Hz | −4.4 dB | 0.58 | **0.47** |
| Watterson 2 ms / 0.2 Hz | −1.6 dB | 0.25 | **0.18** |
| AWGN | all | 0.00 | 0.00 |

Better or equal at every point, never worse. At the 1 Hz fringe that is 21 failures out of 60 down to 12 — expected transmissions per frame 1.54 → 1.25, about **19% more ARQ throughput** where it is needed, for zero airtime and a table lookup per bit. AWGN is unchanged, as it should be: interleaving buys nothing against white noise.

The gain is smaller than one might hope, for a reason worth writing down: at 1 Hz Doppler the fade coherence time is ~0.5 s, so a 13.5 s burst already spans ~26 independent fades and carries intrinsic time diversity. Interleaving helps most when a single fade covers a large fraction of the burst — which is why the 0.2 Hz slow-fade case gains at least as much, and that is the realistic NVIS condition.

### What it did *not* do

It does **not** rescue the deeper-M configurations. M=4 keeps its ~0.40 error floor at −0.2 dB with interleaving in place, so my earlier hypothesis that interleaving would make M=4/M=8 viable is **wrong** — M=32 stays the rung. The M=4 floor is non-monotonic in SNR, so it is not noise-limited; I have not chased it further since the config is not a candidate.

### Two defects found while wiring it up

- `h->llr` was allocated as `code->N` floats, but `mfsk_demod` writes one LLR per transmitted bit slot (`NPAY*bps`). For M=32 both are 1600 so nothing overflowed — M=8 gives 1602 slots and a two-float overrun. Buffers are now sized by `nb`.
- `mfsk_be_close` never freed `h->bb`: `rxcap` complex doubles, ~3.4 MB leaked on every open/close cycle, and the mode pool opens/closes per session.

Unit test covers permutation validity, cross-platform determinism, and the spread property. Full C suite green; `go test -count=1` green (the earlier "ok (cached)" was Go reusing a result — the test binary hadn't changed even though `mercury` had).

**Note this is an on-air bit-ordering change — both ends must run it.** Pre-2.0 has no compatibility constraint and MFSK has not shipped in a release.

rafael2k:
## Frequency search — `ba1dce9`. Plus two ideas measured and rejected.

Read modem73 (public domain) looking for things to borrow. Its MFSK is almost the same waveform as ours — 32 tones, 1000 Hz, 5 bits/symbol, rate 1/2 — but with a *weaker* code (K=7 Viterbi against our LDPC N=1600), so a sensitivity bake-off isn't worth running. Three ideas looked worth having. One was, two weren't.

### 1. Frequency search — landed, and it fixes a real defect

**We had no frequency tolerance whatsoever.** Measured on the pre-fix decoder, clean channel:

| dial offset | before | after |
|---|---|---|
| 12 Hz (0.38 bins) | 10/10 | 10/10 |
| **16 Hz (0.51 bins)** | **0/10** | **10/10** |
| 47 Hz (1.50 bins) | 0/10 | 10/10 |
| 94 Hz (3.01 bins) | 0/10 | 10/10 |
| 110 Hz (3.52 bins) | 0/10 | 0/10 *(past ±3-bin design range)* |

16 Hz — half a subcarrier — and the mode is stone deaf. That is well inside what two HF radios drift apart by. **Every sensitivity number in this PR was measured on a perfectly tuned simulator, so the entire test suite was blind to it**; on air it would have presented as "the robust mode just doesn't work", with nothing in the logs to explain why. Tuning tolerance is now ±12 Hz → ±100 Hz.

Two requirements that weren't obvious:

- **Half-bin steps, not whole.** A whole-bin grid leaves the worst case exactly between two hypotheses, where tone energy splits across two bins and neither template captures it. Measured: 31/62/94 Hz recovered to 10/10, but 16/47/110 Hz stayed at 0/10.
- **Argmax, not first-past-the-gate.** At half-bin spacing a neighbour 0.5 bin off still clears the threshold, so first-past latches the wrong offset — acquisition "succeeds" and every frame then fails CRC. The tell was 94 Hz going 10/10 → 0/10 when the grid got *finer*, which is backwards.

Cost is contained by latching: a CRC pass confirms the offset and later searches try that one hypothesis alone, and a resident payload failing CRC drops the lock so a wrong latch can't wedge the decoder. **+21%** decode CPU with signal present, none idle. No regression: AWGN cliff identical point-for-point at zero offset (20/20, 18/20, 11/20, 0/20 over −9.1..−12.1 dB SNR3k), false alarms still **0/200** despite 13× the hypotheses.

### 2. Goertzel detection — not worth doing

Profiled the decoder: **0.23% of real time.** 0.18 s CPU per 78.7 s of audio — downmix 0.067 s, sync 0.064 s over 78 calls, payload 0.049 s over 6 bursts. On idle noise sync is 0.002 s over 156 calls, because the per-symbol floor rejects instantly.

The old "MFSK RX runs at half real time" was real, but it is already fixed — incremental downmix, search trim, and above all the pattern-ACK gate. Even 20× slower on a Pi that is ~5%. There is nothing left to optimise, so I'm not adding a second tone detector to maintain.

### 3. Smaller frames — rejected, it would make fading *worse*

modem73 uses 4.42 s frames against our 13.12 s, which looked like the obvious answer to our fade behaviour. It isn't.

First, arithmetic: a shorter frame needs a shorter codeword, and all five of our codes are N=1600. `H_256_512_4` in the freedv tree isn't IRA/staircase, so our `{C,Enc}` format would need an encoder derived — real work. Before doing it I tested the premise. FER vs Doppler at fixed SNR (No=−12, 2 ms, 30 trials):

| Doppler | independent fades per burst | FER |
|---|---|---|
| 0.1 Hz | 2.6 | **1.00** |
| 0.2 Hz | 5.2 | 0.30 |
| 0.5 Hz | 13.1 | 0.23 |
| 1.0 Hz | 26.2 | **0.17** |
| 2.0 Hz | 52.5 | 0.20 |

The burst **relies on spanning many independent fades**. Cutting it 3× divides that by 3: at 1 Hz, 26 → 9 fades (FER ~0.17 → ~0.27); at 0.2 Hz, 5.2 → 1.7, which the curve puts at a dead link. Slow NVIS fading is the case that matters, and cheaper retransmits don't come close to offsetting that.

It also retro-explains the interleaving result earlier in this PR: interleaving helped *more* at 0.2 Hz than at 1 Hz precisely because the slow case has so little intrinsic diversity to start with.

modem73 can afford short frames because their MFSK is a backup mode behind an OFDM/ROBUST ladder, and their published NVIS data sits at 0–16 dB SNR where margin is ample.

### Where this leaves the rung

AWGN 50% FER **−11.2 dB**, fading cliff ~**−4.9 dB** with the sync-caused error floor gone, ~19% less retransmission under fade from interleaving, and now usable on a radio that isn't perfectly zero-beat. The remaining ~13 dB to capacity is not reachable by tuning this waveform — that needs coherent, pilot-aided modulation (modem73's ROBUST family territory), which is a separate piece of work.

Still OTA-pending, and note the interleaver and this change are both on-air format changes: **Pedro's run needs matched builds on both stations.**

rafael2k:
Re-ran the sweeps and dug into the 1 KB stall. **It is not a byte-count stall — it is the ladder climb to DATAC1**, and the localisation is now tight enough to hand to whoever fixes it.

## What the stall actually is

Reproduced first try, same 276/1054. With both peers at DEBUG, the receiver's decode log lines up exactly with the sender's ladder climb:

| A transmits | mode | B decodes / ACKs |
|---|---|---|
| 10.7 s | MFSK (13.6 s) | ✓ |
| 26.4 s | DATAC15 (4.40 s) | ✓ |
| 32.9 s | DATAC4 (5.83 s) | ✓ |
| 41.0 s | DATAC3 (3.84 s) | ✓ — `rx_total=276` |
| **47.0 s** | **DATAC1 (4.83 s)** | ✗ nothing |
| 59.9 → 317 s | DATAC1 × 22 retries | ✗ nothing |

276 bytes is simply the sum of the four frames that land before the ladder reaches DATAC1. B then sits silent until `IRS inactivity (181s without RX)` and the session dies. The 102-byte payload used in the reach sweeps completes before the ladder ever gets that far, which is why those cells look clean.

## Three things ruled out

**DATAC1 is not broken.** Pinned to the DATAC1 rung (`MERCURY_PIN_LADDER=4`), the same binary moves the whole 1054 bytes in 88.6 s, PASS, with B decoding DATAC1 frames normally. The mode works; arriving at it by a mode switch is what fails.

**Mode selection is not broken.** I built the branch with a log line in `rx_decoder_bind_mode` — the payload worker binds correctly and holds it across the burst:

```
[+38.498s] INSTR bind_mode 18 -> 12     (DATAC4  -> DATAC3)
[+44.546s] INSTR bind_mode 12 -> 10     (DATAC3  -> DATAC1)   <- correct
[+61.293s] INSTR bind_mode 10 -> 12     (step-down after the stall)
```

A's DATAC1 burst arrives at 47.0 s, squarely inside that window. The decoder is pointed at the right mode, is being fed, and produces nothing — no decode, no flush, no backlog warning.

**The mirror is not broken.** `IRS RX-mode mirror climb to level 4 (mode=10)` fires at 44.485 s and the snapshot carries `peer_tx_mode` through to `select_payload_rx_mode` as intended.

## Where that leaves it

The receiver has the right decoder, bound at the right time, fed the right audio, and does not acquire — only when it reached DATAC1 by switching rather than starting there.

My working hypothesis was the `bind_mode` comment that deliberately skips the sync reset ("stale rxbuf content is harmless"), on the theory that leftover state from a 500 Hz mode mis-scales the preamble normalizer for a 1700 Hz one — the ladder's first big bandwidth jump, and the only failing step. **I could not confirm it, and one fact argues against it:** trunk carries the identical no-UNSYNC code and makes the same DATAC3 → DATAC1 jump successfully on the same bench. So something branch-specific is interacting with it rather than the skip alone being wrong. Testing it directly needs a `set_sync` hook, which the backend vtable does not currently expose.

**Next step for whoever picks this up:** instrument the payload worker's ring occupancy and decode-attempt count across the switch. That distinguishes "the burst never reached the decoder" (ring/backlog) from "it reached it and acquisition failed" (decoder state), which is the one fork the current evidence does not settle.

## Likely relevance to the fading regression

Worth checking whether these are one bug. Fading makes the ladder move far more, so if a climb into DATAC1 can silently kill a session, the 1/5 fading result may be the same defect sampled differently rather than a separate weakness. I have not verified this — flagging it so the fading work does not start from scratch.

## Reproduce

```
MERCURY_BIN=<mfsk-margin binary> MERCURY_CH_ENGINE=ch MERCURY_CH_NO=-14.82 \
MERCURY_TEST_PAYLOAD_KB=1 MERCURY_TEST_VERBOSE=1 \
go test -C tests/integration -count=1 -run 'TestMercuryARQTransfer$' -v
```

`MERCURY_TEST_VERBOSE` is new — the harness previously had no way to run the peers at DEBUG, so the first pass at this had to infer modes from PTT durations. Committed separately on `test-verbose-hook`.

rafael2k:
Follow-up: two more candidate causes tested and **both eliminated**, plus a correction to my own reading.

## HARQ is not the cause

`MERCURY_HARQ=0`, same binary, same channel: still 276/1054, and still **zero** DATAC1 frames decoded.

I initially read this run as "HARQ off decodes one DATAC1 frame", which would have made HARQ the prime suspect. That was wrong — `grep -c "Decoded frame mode=10"` also matches `mode=100`, the MFSK mode, so the one hit was an MFSK decode. Corrected: HARQ-on and HARQ-off are identical here, and the "synced but zero bytes" signature that made HARQ look guilty (it matched a real HARQ bug we fixed earlier this year) is a coincidence.

## The receiver is not starved, and its sync is real

Instrumented the payload worker with a ring/throughput heartbeat and a per-attempt decoder probe:

```
[+47.079s] INSTR hb  mode=10 ring=122B chunks=15 samp=13200
[+49.135s] INSTR dec mode=10 att=18 sync=16 rxst=2
[+51.231s] INSTR dec mode=10 att=19 sync=19 rxst=2     <- synced 19/19 during the burst
[+59.943s] INSTR dec mode=10 att=5  sync=3  rxst=0     <- drops after it ends
```

- **Audio reaches it in real time.** Ring stays at 122–282 B (not backing up) and it consumes ~8360 samp/s against 8000 arriving. The backlog/starvation theory is dead.
- **Sync is genuine, not stuck.** It rises during A's burst and falls when the burst ends, exactly as it should. The control decoder shows `sync=0` throughout the same window, which is the correct negative control.

So the decoder is bound to the right mode, fed on time, and truly acquires the DATAC1 preamble — and no frame ever comes out. `rxst=2` is SYNC without BITS: it locks and the frame never completes.

## What I could not settle, and why

I ran the working case (`MERCURY_PIN_LADDER=4`) through the same instrumentation to diff against the failing one. They look **the same** at the sample points:

```
pinned  (works): sync=16..19  rxst=2
climbed (fails): sync=16..19  rxst=2
```

My probe samples every 2 s, which is far coarser than a decode event, so it captures the mid-burst state in both runs and misses the moment that differs. The instrumentation cannot separate them — that is a limitation of what I built, not evidence that they are identical.

**Concrete next step:** log *every* DATAC1 decode attempt (unsampled) with `nin`, `demod_count` and the full `rx_status` bits, for a pinned run and a climbed run, and diff the two traces around the first burst. That will show whether the climbed decoder is being fed misaligned sample counts after the switch, or whether the LDPC is simply failing CRC on good sync.

## Standing summary

| candidate | verdict |
|---|---|
| byte-count / buffer limit at 276 | **no** — 276 is just the four frames below DATAC1 |
| DATAC1 mode broken | **no** — pinned delivers 1054/1054 in 88 s |
| mode selection / mirror | **no** — binds to 10 at 44.5 s and holds it |
| ring starvation / backlog | **no** — ring near-empty, real-time consumption |
| acquisition failure | **no** — sync 19/19 during the burst |
| HARQ | **no** — identical with `MERCURY_HARQ=0` |
| frame completion after good sync | **open — this is where it is** |

Reproduce with `MERCURY_TEST_VERBOSE=1` (now on `test-verbose-hook`) and the command in my previous comment.

rafael2k:
> [!WARNING]
> **Correction (2026-08-19):** the "fails on 72 of 100 seeds on trunk" claim below is wrong. Trunk passes **100/100** seeds at its own 2-hour budget; that number was measured on this branch's rewritten test, not on trunk. Measured correction: https://github.com/Rhizomatica/mercury/pull/160#issuecomment-5345083876

## Blocker 1 (the 1 KB stall) is root-caused and fixed

The description above lists "stalls at exactly 276/1054" as an open blocker. It is closed, and the cause was not where I was looking.

The retained frame is immutable — it is never re-framed smaller, because the seq<->bytes identity has to stay fixed for a duplicate to be idempotent on the peer. So **the width it is read at permanently decides which modes can ever carry it.** Frame seq=4 was read at DATAC1's full 502 bytes while *probing* DATAC1, and no lower rung's slot holds 502 bytes, so `mode_that_fits()` pinned every retransmission to DATAC1 while the ladder stepped 4->3->2->1->0 beneath it.

The proof is in the keydowns. Every transmission after the probe is **4.817 s**, DATAC1's frame duration, while the ladder claimed MFSK (13.5 s):

```
+45.9s  Ladder step-up   to 4        <- probes DATAC1
+59.7s  Ladder step-down to 3 (retry)      keydown 4.817 s
+72.6s  Ladder step-down to 2 (retry)      keydown 4.817 s
+85.4s  Ladder step-down to 1 (retry)      keydown 4.817 s
+98.3s  Ladder step-down to 0 (retry)      keydown 4.817 s
```

Those step-downs never reached the air. What they did move was the peer's RX mirror, which follows the same delivery-driven rule — so it walked away from the one mode actually being transmitted. Only one payload decoder runs at a time, so both ends then sat at the MFSK floor **in perfect agreement for 120 s** while DATAC1 was on the air.

Fix: a fresh read is capped so the frame still fits some rung *below* the current one. A failed probe then costs one retry instead of the session.

## Measured (make-clean builds, ch AWGN, Go harness)

| | pre-fix branch | now |
|---|---|---|
| 0 dB, 1054 B | 276/1054, wedged, 3/3 | **952/1054**, still delivering at the deadline |
| fringe -12 dB, 102 B | 50.44 s | 50.45 s |
| clean, 1054 B | 81.18 s | 87.10 s (**7% slower**) |

The 7% is the cost of the cap: one frame per rung climbed goes out narrower than its rung could carry. It buys immunity to a wedge that costs the entire session.

## Two ideas measured and rejected

Both looked obviously right and both are recorded in the commits so they are not retried:

- **Start the ladder above the MFSK floor.** A session only exists because DATAC16 was carried both ways, and MFSK sits 10-12 dB below that. Worth 81.18 -> 71.57 s on a clean link; costs **2x** at the fringe (50.44 -> 101.71 s).
- **Climb two rungs per clean delivery.** Worth 81.18 -> 72.08 s clean, and free at the fringe once the floor is excluded from the jump. Collapses the 0 dB cell to **202/1054** — it races past what a marginal channel carries and pays an ACK timeout per overshoot.

Faster on the channel that never needs it, far worse on the one that does. Neither is in the branch.

## Two infrastructure bugs that faked product bugs

Both cost real time here and are fixed in `410f5b1`:

1. **No header dependency tracking anywhere in the main build.** Objects come from make's implicit rule and no Makefile emitted or included `.d` files. Editing a header rebuilt nothing. It produced a bench number 18% off, and a binary with half the program compiled against each of two struct layouts — which froze deterministically and which I briefly diagnosed as a real deadlock.
2. **`go test` orphans its children on `-timeout`.** It calls `os.Exit`, so cleanups never run; the orphans keep their audio threads spinning and starve the harness's real-time pacing until runs stop completing, presenting as a CALL sent once and never retried.

## Blocker 2 (fading) needs re-reading, not re-running

The fading numbers in the description came from `test_sim_fade_cliff_downgrades`, which asserted full delivery and **fails on 72 of 100 seeds on trunk**. It is green in CI because seed 42 is lucky. It now asserts prefix integrity and takes `SIMSEED` so the rate is measured by sweeping rather than sampled by one run. The 1/5-vs-5/5 figures above should be treated as unmeasured until re-swept.

Details in `docs/ARQ-FRAME-SIZING.md`.

## Still open

A frame read at a rung that *has* delivered can still be stranded if the band collapses afterwards (42-43/100 on the sweep, unchanged before and after). Closing that needs frame identity that survives re-framing — a stream offset rather than a bare sequence number. Wire change, pre-2.0 permits it, and it is most of what multi-frame bursts need.

rafael2k:
## Correction: the "fails on 72 of 100 seeds on trunk" claim was wrong

In [this comment](https://github.com/Rhizomatica/mercury/pull/160#issuecomment-5321332045) I wrote that `test_sim_fade_cliff_downgrades` "asserted full delivery and **fails on 72 of 100 seeds on trunk**". That is not true, and the docs in this branch repeat it. I've now measured it properly.

The baseline arm in that sweep was never trunk — it was *this branch's rewritten test running on this branch's rewritten FSM*. I labelled a branch number as a trunk number.

### What the sweep actually shows

Method: took trunk (`f79a28e`) and parameterised only the seed and the time budget of its own `test_sim_fade_cliff_downgrades`, changing nothing else. This branch already accepts `SIMSEED`. 100 seeds each.

| arm | budget | assertion | pass |
|---|---|---|---|
| **trunk, as it ships** | 2 h | full delivery | **100 / 100** |
| trunk, budget cut to 1 h | 1 h | full delivery | 12 / 100 |
| **this branch, as it stands** | 1 h | prefix integrity | **59 / 100** |

So trunk is **not** a seed lottery — it passes every seed I tried. The lottery appears only when the budget is cut to one hour, which is precisely what trunk's own comment warned about:

> *At one hour this test was fitted to seed 42 … At two hours seeds 42-45 all pass.*

That two-hour budget is load-bearing, and trunk had already paid for the lesson.

### What this branch's version does

It changes two things at once: the budget (2 h → 1 h) and the assertion (full delivery → prefix integrity). The budget cut is reasoned in the comment — the band is restored to 12 dB so the remainder drains at a fast rung rather than at the floor's ~22 B / ~11 s — so it is not simply a re-break of what trunk fixed, which is the other thing I had assumed and which is also wrong.

But it has its own seed sensitivity, 41 failures in 100. Classified by cause:

| failure | count |
|---|---|
| "did not recover to a fast mode after the fade" | 31 |
| "did not climb on the good band" | 10 |
| **prefix integrity (correctness)** | **0** |

That split is the useful part. **Zero integrity failures across all 100 seeds** — the property the test was rewritten to protect, that no re-framing can duplicate or reorder delivered bytes, holds on every draw. What is seed-sensitive is the two new ladder assertions (`max_level >= 3`), which are fitted the same way the old full-delivery assertion was.

### Consequences

- The claim that trunk's test carried no information is withdrawn. It carries information; I measured the wrong binary.
- The `48/100 fail` and `42-43/100` numbers in `docs/ARQ-FRAME-SIZING.md` are branch-vs-branch. They say nothing about trunk and the doc should not imply otherwise. Fixing that wording next.
- The ladder assertions here need the same treatment the delivery assertion got — either a budget with margin, or a sweep-based threshold instead of a fitted one — otherwise this test will misread the next timing change as a regression, which is the exact trap it was rewritten to escape.

None of this changes the frame-sizing fix itself, which is measured on the bench cells, not on this test. It does mean this branch still has no measurement against trunk — that's the outstanding item before #160 can be judged.

rafael2k:
## Measured against trunk at last

This branch had never been benchmarked against trunk — every A/B in it was branch-vs-branch, which is what produced the retracted seed claim above. Here is the comparison.

**Method.** Arms are `make clean` full builds of whole commits: trunk `f79a28e` (md5 `cd3f0573…`) and this branch `aaac494` (md5 `69f9a905…`). Both run under **one harness**, not each branch's own — this branch's `TestMercuryARQTransfer` asserts `"handshake completed on connect confirm (pattern)"`, a branch-only feature that trunk fails for reasons unrelated to speed, so comparing each branch's own test would have repeated exactly the mistake being corrected. I took this branch's harness (it carries the orphan-reaping `spawn_linux.go` fix), removed the pattern and mode assertions, and made it report `delivered/want/seconds` instead of hard-failing, so a stalled arm still yields a number. `ch.c` is byte-identical between the two branches, so the channel is not a confound. Arms interleaved, children reaped by argv between every cell. Timing covers payload-write → last byte delivered, excluding connect.

| cell | arm | delivered | median | spread | n |
|---|---|---|---|---|---|
| clean, 5 KB | trunk | 5134 / 5134 | 94.8 s | 24.4 % | 5 |
| | **this branch** | 5134 / 5134 | **77.3 s** | **0.4 %** | 5 |
| 0 dB (`No=-14.82`), 1 KB | **trunk** | **1054 / 1054** | **189.3 s** | 0.1 % | 2 |
| | this branch | 952 / 1054 | **never completes** | 0.1 % | 2 |
| fringe (`No=-2.82`, ≈ −12 dB SNR3k), 102 B | trunk | **0 / 102** | — | — | 2 |
| | **this branch** | **102 / 102** | **22.7 s** | 0.1 % | 2 |

Every arm-cell reproduced to within 0.4 %. These are effects, not draws.

### The fringe cell is the case for this branch

Trunk delivers **zero bytes in 315 s** at −12 dB, twice. This branch delivers all 102 in 22.7 s, twice. That is not a speedup — it is the difference between a link and no link, and it is what the MFSK floor was built for.

### The 0 dB cell is the case against it, and it inverts the PR's headline

The "276 wedged → 912/1054 still delivering" result in `docs/ARQ-FRAME-SIZING.md` was measured against this branch's own pre-fix build, so it read as a rescue. Against trunk it reads differently: **trunk was never wedged at 0 dB.** It completes all 1054 bytes in 189 s, twice, while this branch stops at exactly 952 both times and never finishes.

952 = 28 × 34 against 1054 = 31 × 34 — the branch is precisely **three frames short**, repeatably. That is the signature of the residual stranded-retained-frame defect the doc already lists as unfixed ("a frame read at a rung that HAS delivered is still stranded if the band collapses afterwards"). At 0 dB that residual is worse than trunk's behaviour, not better.

### The clean cell went the other way from what the doc predicts

The doc claims the cap costs ~7 % on a clean channel (81.18 → 87.09 s). Against trunk it is a **~18 % win**: 77.3 s vs 94.8 s median. That 7 % was also a branch-vs-branch number.

Worth flagging separately: **trunk's clean cell is unstable** — 76.3 / 88.3 / 94.8 / 94.9 / 95.0 s, a 24 % spread, and it looks bimodal rather than noisy. This branch is pinned at 77.2–77.5 s across five runs. Whatever causes trunk's slow mode is on trunk and is worth a look independent of this PR.

### Where that leaves #160

The arms trade; they do not rank. This branch buys the fringe outright and is faster and far more deterministic on a clean channel; it currently gives up completion at 0 dB, in a known and localised way.

The 0 dB stall is the one blocker, and it is the residual already identified — offset-based frame identity instead of a bare `tx_seq`, which makes re-framing idempotent by construction. Until that lands, merging trades a mid-SNR regression for fringe reach.

**Limits of this measurement, stated plainly:** `ch` is noise-only, so none of these cells include fading; the two decisive cells are n=2 (though both reproduced to 0.1 %); and this is one host, so the absolute times are not portable — only the arm-to-arm comparison is.

rafael2k:
## The 0 dB stall is fixed — and it was neither of the causes this PR documents

Pushed `2b22ff3`. The 0 dB cell now completes; the fringe cell is unchanged.

| cell | before | after |
|---|---|---|
| 0 dB (`No=-14.82`), 1 KB | 952 / 1054, **ten consecutive runs**, never completes | **1054 / 1054 ×3** |
| fringe (`No=-2.82`, ≈ −12 dB), 102 B | 102 / 102 in 51 s ×2 | **102 / 102 in 51 s ×2 — identical** |

### It was never a wedge

The doc calls this a stranded retained frame. It is not, and the giveaway was sitting in the data the whole time: **952 exactly, every single run.** A channel-loss outcome does not land on the same byte ten times. The transfer is progressing the entire time at a steady rate, and the harness deadline truncates it at the same place.

So the question was never "why does it wedge" but "why is it ~5× too slow". The airtime answers it — 12 of 21 bursts went out on **MFSK** (90 B per 13.5 s burst) on a channel that carries **DATAC3** (118 B per 3.82 s), which is what trunk uses to finish this same transfer in 189 s.

(I also reported mid-investigation that the reverse path was failing. That was wrong: the sender received **all 12** ACKs the receiver sent. The `snr_local≈4.7 dB` readings that suggested a healthy forward path were survivorship bias — the receiver only logs SNR for frames it actually decoded. Counting bursts instead of SNRs: 23 sent, 12 decoded. The loss is forward.)

### Two defects, and neither is fixed by the other

**1. The two ladders diverge under forward loss.** A lost burst is *invisible* to the receiver — the retransmission is the first copy it ever sees, so it scores a clean delivery and **climbs**, while the sender scored a retry and **descended**. With a single payload decoder that split is total deafness, which loses the next burst and widens it. The receiver's mirror was measured oscillating `0 → 1 → 0` against a sender parked at the floor.

Fixed with `ARQ_FLAG_RETX` on a spare flags bit. `tx_frame_retx` already decided `clean` on the sending side — it was simply never put on the wire.

**2. `last_good` ratchets down.** A miss at a rung that had delivered ran `(*level)--; *last_good = *level;`, which **erases the evidence that the rung works**, so the next miss lowers it again. Ordinary loss walks a healthy link to the floor and anchors it there. Abandoning a *proven* rung now requires `ARQ_RETRY_DOWNGRADE_THRESHOLD` consecutive misses — a knob that already existed, already defaulted to 2, and was already logged at startup, but which nothing read.

They only work together, which is why testing them separately mattered:

| arm | 0 dB, ×3 reps |
|---|---|
| baseline | 952 / 1054 |
| **RETX only** | **726 / 1054 — worse than baseline** |
| both | **1054 / 1054** |

RETX alone keeps the ladders in lockstep, but locked onto the floor the ratchet drags them to.

### One rejected fix, measured

My first attempt gated *every* descent behind the threshold, including the fallback from an overshot probe. That is much worse than the bug: the session cannot escape a rung the channel will not carry, burns `seq=4`'s ten retries and **disconnects at 248/1054**. The overshoot fallback is deliberately left immediate, and that is also why the fringe is unaffected — the descent to the floor at −12 dB is almost entirely overshoot.

### Guards

Two deterministic tests, no wall-clock timing:
- `test_proven_rung_survives_a_single_miss` — pins both rules, including that `last_good` is not lowered before the rung is actually abandoned.
- `test_retx_flagged_frame_does_not_climb_the_mirror` — pins the ladder symmetry.

172/172 green, clean build, no new warnings.

### Note for review

`ARQ_FLAG_RETX` is a **wire change** (a previously-spare bit in the flags byte). Pre-2.0 permits it, but both ends must run it: a sender with the flag talking to a receiver without it is no worse than today, while the reverse gains nothing.

Method as before: one harness for both arms, binaries pinned by path and checked present before each run (an earlier attempt silently measured a binary that had never been copied, and another collided with a still-running rep — both discarded).

rafael2k:
## Blocker found under fading: this branch cannot CONNECT where trunk can

I ran the Watterson fading model against this branch. The ladder fixes hold up — but the fading runs exposed something the `ch` cells could never have caught, because `ch` is noise-only.

**Under ITU-R `good` fading (1 KB payload), CONNECT:**

| `No` | trunk `f79a28e` | this branch |
|---|---|---|
| −30 | — | connects, 1054/1054 |
| **−25** | **connects, 1054/1054** | **never connects** |
| −20 | connects, 1054/1054 | — |
| −15 | connects, 536/1054 (deadline) | — |

Trunk's connect survives **at least 10 dB deeper** than this branch's. At −25 the answerer decodes *nothing at all* — no CALL in 58 s — while trunk at the same setting completes the whole transfer.

**This is not caused by today's ladder fixes.** The pre-fix branch build fails identically at −25, so it belongs to the connect work already on this branch.

That is the wrong way round for a branch whose entire purpose is reach. The MFSK fringe rung cannot help if the session never establishes: at −25 trunk moves 1054 bytes and this branch moves zero.

**Where I'd look.** The CALL/ACCEPT frame itself is unchanged (14 bytes, DATAC16, byte-identical to trunk), so this is not frame robustness. The connect *timing* did change here:

- `ARQ_CONNECT_GRACE_SLOTS` (2 extra wait slots for ACCEPT) removed
- `ARQ_CONNECT_BUSY_EXT_S` removed
- `ARQ_ACCEPT_RX_WINDOW_MS` changed from a fixed 9000 ms to a computed window
- the CALL retry clock re-anchored to when the burst leaves the air rather than when it was queued

The last one lengthens the effective retry interval by roughly the CALL's own airtime. On a flat channel that is free — it was measured as −3.1 s on connect. **Under fading it is not free**: fewer CALL attempts per unit time means fewer independent chances to land one in a fade null. That is the same "airtime is diversity" lesson this PR already documents for the ACK pattern, applied to the connect. I want to stress this is a *hypothesis* consistent with the measurement, not something I have isolated — the measurement is that the answerer hears nothing.

**Caveat on the axis.** These dB are not trustworthy in absolute terms — the `SNR3k = −No − 14.82` calibration was established without multipath, and this model shows other implausible behaviour (full delivery at −30, no connect at −25, a 5 dB step). Treat the numbers as a **relative** comparison at identical settings. The relative result is unambiguous and reproduces: same harness, same channel, same settings, trunk connects and this branch does not.

**What this means for #160.** The 0 dB stall is fixed and the fringe is intact, but I would not merge until the connect gap is understood. The bench cells I used to benchmark this branch cannot see it, which is precisely why it went unnoticed.

Worth checking on air: whether a marginal real link shows the same asymmetry.

rafael2k:
## Connect regression: isolated to the ACCEPT decode, with a correction

Followed this up with per-direction noise. The result is sharper than my previous comment, and one thing I said there was wrong.

### Correction

I wrote that at −25 "the answerer decodes nothing at all — no CALL in 58 s". **That is wrong.** It was an artefact of reading an INFO-level log that hid the DEBUG lines. With `MERCURY_TEST_VERBOSE=1` the answerer clearly *does* receive the CALL:

```
[DBG] [modem-rx] Decoded frame mode=23 (DATAC16) bytes=16 snr=12.22
[DBG] [arq-fsm] state=ACCEPTING dflow=IDLE_ISS ev=RX_CALL
```

It goes to ACCEPTING and transmits ACCEPT twice. The forward path is fine.

### What actually fails

The **caller never decodes the ACCEPT**. Per-direction noise isolates it completely:

| forward | reverse | trunk | this branch |
|---|---|---|---|
| −25 | clean | — | **connects, 1054/1054** |
| clean | −25 | **connects, 1054/1054** | **never connects** |

Same harness, same channel, same fading preset, only the direction of the noise changed. The branch's caller cannot decode a DATAC16 ACCEPT on a faded reverse path where trunk's caller can.

### What it is not

Ruled out by measurement rather than by reasoning:

- **Not the CALL** — the answerer receives it and responds.
- **Not the ACCEPT RX window.** Aligning both logs on wall clock, ACCEPT#1 is on air `50.038–53.755` while the caller is listening `49.478–57.489`. It falls *entirely inside* the caller's gap between CALL bursts and is still not decoded. (ACCEPT#2 does collide with the caller's own CALL#4, but #1 does not.)
- **Not HARQ.** Trunk still connects at `rev=-25` with `MERCURY_HARQ=0`, so soft-combining is not what is saving it.
- **Not the frame.** CALL/ACCEPT is byte-identical to trunk — 14 bytes, DATAC16.
- **Not ring backlog or decoder starvation.** No flush/backlog/overrun evidence in the caller's log, and its DATAC16 and MFSK decoder workers are both running.
- **Not the ladder fixes in `2b22ff3`.** The pre-fix branch build fails identically.

So it is a **receive-sensitivity difference on control frames under fading**, somewhere in this branch's RX path, and it is graded rather than a hard bug: the branch connects at `-30` and fails at `-25`, while trunk is still connecting at `-15`.

### Where I would look next

A bisect over this branch's RX/decoder commits is the obvious next step — the dual-decoder wiring, the mode-aware backlog cap added so a 13.5 s MFSK burst is not chopped, and the postamble-anchored MFSK sync fallback are the changes that touch the shared capture path, and any of them could cost control-frame sensitivity without showing up on a noise-only channel.

Caveat unchanged: the absolute dB from this model are not trustworthy (see the connect/data inversion in my previous comment). The finding is the **relative** trunk-vs-branch gap at identical settings, and it reproduces every time.

rafael2k:
## Merged current trunk (1.9.13); connect regression reproduced, one suspect eliminated

### Branch is current

Merged `mercuryv2` at 1.9.13 — 109 commits — so this is now **67 ahead, 0 behind**. Merged rather than rebased on purpose: 66 commits over a trunk that rewrote the same RX-sizing code means re-resolving the identical `modem.c` collision at nearly every step.

Five conflicts, mostly "both sides fixed this, keep the better one":

- **`modem/modem.c` → trunk.** Trunk had independently generalised *this branch's own* RX-sizing fix into `rx_burst_capacity_samples()`, identical formula (longest burst + 3 s guard, 2 s floor). Keeping ours would not have compiled — `RX_WORKER_RING_BYTES`/`RX_MAX_BACKLOG_SAMPLES` were removed by that refactor.
- **`arq_protocol.c` → ours.** Both define `arq_protocol_longest_burst_s()`; ours scans the ladder (correct — only ladder rungs carry payload bursts) and falls back to MFSK's 13.5 s.
- **`arq_fsm.c` / `test_arq_fsm.c` → ours.** Trunk's additions reference `ARQ_EV_RX_{TURN,MODE,KEEPALIVE}*` and `tx_window_count`, all removed by the 21→11 state rewrite.
- **`tests/Makefile`** → union minus `test_arq_olla`, whose source this branch deleted when OLLA became the delivery-driven ladder.

**One defect git merged cleanly and silently, worth flagging to anyone else merging trunk into an ARQ branch.** Trunk hardened the session-ID gate so *"zero is not a wildcard"*. This branch posts its in-session ACK with `session_id = 0`, because that ACK is a bare Welch-Costas pattern with no header to put an ID in. Merged naively it compiles, reads fine, and **silently discards every in-session ACK** — the data plane stalls with no error anywhere. Reconciled so zero stays non-wildcard for framed events and is accepted only for an ACK, pinned by `test_connected_accepts_zero_session_pattern_ack`. That test earns its place: with the gate trunk-strict it fails, along with two existing tests that also depend on the ACK path.

Unit suite and the Go integration harness both green.

### The connect regression survives the merge

Same experiment as before, now with both builds on the same trunk base. Clean forward (−40), faded+noisy reverse, ITU-R **`good`**, 2 reps/cell:

| rev `--No` | SNR3k | trunk | this branch |
|---|---|---|---|
| −30 | 15.2 dB | 2/2 | 2/2 |
| −25 | 10.2 dB | 2/2 | **0/2** |
| −20 | 5.2 dB | 2/2 | **0/2** |
| −15 | 0.2 dB | 2/2 | **0/2** |

Reproduces the earlier finding exactly: branch connects at −30, fails from −25 down; trunk still connecting at 0.2 dB SNR. Failure mode is `link disconnected before CONNECT completed`.

### The mode-aware backlog cap is eliminated as the cause

This is what the merge bought. The cap was one of the three suspects named above — and the merge **replaced this branch's cap with trunk's `rx_burst_capacity_samples()`**, so both builds now compute it identically. The gap persists unchanged. It is not the cap.

Remaining named suspects: the dual-decoder wiring and the postamble-anchored MFSK sync fallback. Bisect over the branch's 40 RX/decoder commits is running.

### Two method notes, because both nearly produced a false all-clear

**`good` is the HARDEST preset here, not the easiest.** It is 0.1 Hz Doppler / 0.5 ms / 1 path — benign in *delay spread*, which is what the name refers to, but a 3.74 s DATAC16 frame spans only ~0.37 fade cycles, so it sits inside a single fade with nothing to average over. `moderate` (1.0 Hz) spans ~3.7. By this PR's own airtime-is-diversity rule that makes `good` the worst case for acquisition. **The regression reproduces under `good` and does not reproduce under `moderate`** — I first swept `moderate` because it sounded harsher, and got a clean result.

**My first sweep was measured wrong.** I counted "did the output contain `Connected to`", but Go only dumps child logs when a test *fails*. Trunk necessarily fails this branch's pattern-confirm assertion — which fires *after* connect — so trunk scored as connected and the branch scored 0/2 everywhere, including at an easy 15 dB. It contradicted a run I had just watched pass, which is what gave it away. The table above uses the test's distinct pre-connect failure messages instead, which is valid for both builds.

Absolute dB from this model remain untrustworthy; the finding is the relative trunk-vs-branch gap at identical settings, and it reproduces every time.

rafael2k:
## The connect failure is NOT a regression this branch introduced

Bisected it. The headline: **the trunk this branch forked from fails the same way**, so "branch vs current trunk" was comparing two things at once.

All cells: clean forward (−40), reverse −25, ITU-R `good`, same harness, only the binary varies.

| build | what it is | connected |
|---|---|---|
| `bf7d546` | **the branch's actual parent** (old trunk, no MFSK code at all) | **0/4** |
| `904ee4d`, `a444cfc` | later trunk | 1/1, 1/1 |
| `6dec3b0` | the merge-base | **5/5** |
| `mercuryv2` | current trunk | 2/2 |
| `58912ab` | branch at its own trunk merge | 0/1 |
| `0cfb518` | branch, all trunk merged to 1.9.13 | 0/4 |

So: old trunk was broken, trunk fixed it somewhere in `bf7d546..6dec3b0`, and **this branch contains that range yet still fails**. The branch is not missing the fix — it is negating it, or failing for a second reason that looks identical.

That also explains why the earlier bisect attempt was parked: bisecting branch commits is *confounded*, because every commit before `58912ab` sits on the old, already-broken trunk. Their BAD verdicts say nothing about the branch. Only `58912ab` and later are meaningful, and `58912ab` is already BAD.

### Eliminated, with evidence

- **Mode-aware backlog cap.** The trunk merge replaced this branch's cap with trunk's `rx_burst_capacity_samples()`, so both builds now compute it identically. Gap unchanged.
- **ACCEPT RX window (9000 → 18100 ms, `7dab3a8`).** It has a clean mechanism — doubling the window halves the ACCEPT retransmit rate, and at 0.1 Hz the caller needs several independent ACCEPTs — but `642d08d` is already BAD, so the regression predates it. I nearly implemented this fix before bisecting; it would have been wasted.
- **Lock structure.** The `58912ab` merge title ("vtable over per-instance locks") made it a suspect, but the branch preserves `modem_inst_lock[]`, just keyed by `void *ctx` instead of `struct freedv *`.

### Not isolated

The trunk-side fix. `bf7d546..6dec3b0` is 55 files and ~4150 insertions — audioio, resampler, arq_fsm, arq_protocol and more. I initially read the 12-commit log as the content and guessed a single lock fix; that was wrong and I withdraw it. **Bisecting that trunk range to find what actually fixed it** is the concrete next step: once it is known, the question becomes the narrow, answerable "what does this branch do differently to *that* code".

### Method notes for anyone re-running this

- **The test is deterministic here** — the base is 5/5 and old trunk 0/4 — so single-rep cells are meaningful, but confirm any headline claim with repeats. I did.
- **`good` is the HARDEST preset**, not the easiest: 0.1 Hz Doppler / 0.5 ms / 1 path, so a 3.74 s DATAC16 frame spans ~0.37 fade cycles with nothing to average over, where `moderate` spans ~3.7. The failure reproduces under `good` and **not** under `moderate`.
- **Do not score connect by grepping for `Connected to`.** Go only dumps child logs when a test *fails*, and trunk necessarily fails this branch's pattern-confirm assertion (which fires *after* connect) — so trunk scores as connected and the branch scores 0/2 even at 15 dB SNR. Key on the distinct pre-connect messages instead.
- **Vary only the binary.** Build at `<sha>`, then run the *current* harness against it via `MERCURY_BIN`, so older commits' differing assertions do not matter. Each build needs `make clean` — stale objects from a newer tree fail the link with a bare `collect2` error that reads like a genuine break.
- Several commits in the ARQ-rewrite series (`5ef3ef8`, `7f7215f`, `8e673bd`) do not compile standalone, so they cannot be bisected.

rafael2k:
## Isolated the trunk-side fix — and it is *not* what this branch is missing

Following on from the last comment, I bisected the trunk range. The fix is a single commit, and `bf7d546` is its direct parent, so there is no ambiguity:

| commit | verdict |
|---|---|
| `bf7d546` (parent) | **BAD 0/4** |
| `c70fdd6` **arq/fsm: anchor the CALL retry to PTT-OFF, not to enqueue** | **GOOD** |

Its own commit message describes the mechanism exactly: anchored to *enqueue*, the retry fires while the ACCEPT is arriving, so "the retry keys the transmitter on top of the reply it was waiting for, on essentially every connect."

**This branch already contains that fix, character-for-character**, with the same 8 s interval (DATAC16's `retry_interval_s` row is identical in both tables). So the branch is not missing it.

## So the branch fails for a different reason: pure DATAC16 receive sensitivity

Side-by-side run, same channel settings, `MERCURY_TEST_VERBOSE=1`, caller (A):

| | trunk | this branch |
|---|---|---|
| CALL #1 | PTT ON +0.130 → OFF +3.840 | ON +0.131 → OFF +3.842 |
| retry re-anchor | +8.00 s | +8.00 s |
| CALL #2 | ON +11.855 → OFF +15.565 | ON +11.875 → OFF +15.585 |
| **ACCEPT** | **+20.597 `Decoded frame mode=23 (DATAC16) snr=10.87` → CONNECTED** | **never decoded**; retries to exhaustion |

The CALL cadence is identical to within ~20 ms. The caller is not transmitting over the ACCEPT, and there is no flush/backlog/overrun anywhere in its log. At the moment trunk decodes the ACCEPT at snr ≈ 10.9, this branch decodes nothing — and never does for the whole run. Trunk logs 21 `modem-rx` events, the branch 8 (and the branch's 8 are all the *answerer* reading CALLs on the clean forward path at snr 14.8–17.1, so the control plane works in general).

**It is a receive-sensitivity deficit on DATAC16 over a faded path, in the caller, and nothing to do with connect timing.** That confirms the original reading in this PR and rules out the whole timing/collision family.

## Eliminated, each by measurement

- **Mode-aware backlog cap** — the trunk merge equalised it (`rx_burst_capacity_samples()`); gap unchanged.
- **ACCEPT RX window** (9000→18100 ms) — `642d08d` is already BAD, so it predates it.
- **Lock structure** — branch preserves `modem_inst_lock[]`, keyed by `void *ctx`.
- **CALL-retry anchor and interval** — present and identical, and the cadence above proves it at runtime.
- **Pattern correlator starving the RX loop** — gated off in CALLING (`expect_pattern_ack`), which is precisely what that gate was added for.
- **MFSK occupying the payload plane** — `MERCURY_PIN_LADDER=1` (payload pinned to DATAC15, MFSK out of the picture): still **0/3**.
- **HARQ on control frames** — both builds exclude it: `harq_enabled() && mode != FREEDV_MODE_DATAC16`.

## Where I would look next

The remaining structural difference in the caller's RX is the `modem_backend` vtable/pool indirection and how the two-plane dispatcher feeds the control worker. The targeted instrument is to log, on both builds, the per-plane samples-in and the decoder's `nin` consumption around the moment the ACCEPT is on air — i.e. answer "does the branch's DATAC16 decoder actually see the same samples trunk's does?" before theorising further. Everything upstream of that (timing, gating, mode selection, HARQ) is now measured and excluded.

rafael2k:
## Connect fixed — and here is where this branch actually stands on performance

### The connect failure is fixed (`21c01f3`)

It was never receive sensitivity. The **answerer was replying at the wrong moment**:

```
+3.617   B decodes CALL#1
+4.349   B sends ACCEPT#1            (one channel guard later — correct)
+15.361  B decodes CALL#2, logs it, AND DOES NOTHING
+26.161  B re-sends ACCEPT#2         (18.1 s after ACCEPT#1 PTT-OFF)
+23.617..+27.327   A is transmitting CALL#3
```

so the retransmitted ACCEPT landed inside the caller's own burst, where it is deaf, every time.

`fsm_accepting` handled `RX_CALL` by resetting the retry counter and nothing else, leaving the deadline armed at ACCEPT `TX_COMPLETE` to run its course. That deadline is `ARQ_ACCEPT_RX_WINDOW_MS`, sized to hold the window open for the caller's first DATA burst — ~18 s once MFSK is the floor, against 9 s when it was derived from DATAC15. But a CALL just arrived, which is proof the caller is still in CALLING and has *not* begun a data burst. The window answers "how long might a data burst be?" when the evidence says "there is no data burst" — and 18 s against the caller's 8 s CALL interval guarantees the drift above.

Fix: re-anchor the retry to one channel guard from the CALL we just heard — the same schedule `fsm_listening` uses for the first ACCEPT, and for the same reason. `test_accepting_reanchors_accept_retry_on_repeated_call` pins it (19100 ms vs the required 5700 ms).

| rev `--No` | SNR3k | before | after | trunk |
|---|---|---|---|---|
| −30 | 15.2 | 2/2 | 2/2 | 2/2 |
| −25 | 10.2 | **0/4** | **2/2** | 2/2 |
| −20 | 5.2 | **0/2** | **2/2** | 2/2 |
| −15 | 0.2 | **0/2** | **2/2** | 2/2 |

I also have to correct myself: I earlier "eliminated" the ACCEPT window on the strength of `642d08d` being BAD. That cell was **confounded** — it sits on the old trunk that was already broken for the CALLING-side reason. The elimination was invalid, and the window was implicated after all, just not in the way I first guessed.

### Performance: the branch wins at the fringe and loses on good links

Symmetric ITU-R `good` fading, same harness, only the binary varies. Times are wall-clock for the whole transfer including connect. Repeats agree to ~1 %, so these are not noise.

**Fringe — 102 B payload:**

| SNR3k | trunk | branch | |
|---|---|---|---|
| 0.2 dB | 116 s | **98 s** | 1.2× faster |
| −2.8 dB | 146 s | **95 s** | 1.5× faster |
| −4.8 dB | 245 s, 217 s | **95 s, 98 s** | **~2.4× faster** |

The branch is **flat at ~95–98 s across the whole range** while trunk degrades steeply. That is the MFSK floor doing exactly what it was built for: a rung that just works, instead of thrashing on modes that don't.

**Good links — 1 KB payload:**

| SNR3k | trunk | branch | |
|---|---|---|---|
| 10.2 dB | 83 s | 144 s | **1.7× slower** |
| 5.2 dB | 100/100/101 s | 234/234/233 s | **2.3× slower** |

That is a real cost, and it is deterministic. It is the price of opening the ladder at the MFSK floor and climbing: on a channel that could carry DATAC1 immediately, the session spends its early life in 13.5 s / 98 B bursts.

**Neither build delivers 1 KB below ~5 dB within 300 s** — both time out at 0.2 dB and lower. So the fringe advantage above is about *getting small payloads through at all*, not bulk throughput down there.

### What I would take from this

The crossover sits around 0–5 dB SNR. Below it the branch is clearly better and the reason to have built it; above it, opening at the MFSK floor costs roughly a factor of two. `12ed10a` ("keep the ladder starting at the MFSK floor — measured, not assumed") is worth revisiting with these numbers: a faster climb, or a start rung chosen from the connect SNR estimate, would recover the good-link case without giving up the fringe.

Caveats: one channel model, one host, and the absolute dB from watterson are not trustworthy — the comparisons are relative, at identical settings, and reproduce.

rafael2k:
## Ladder fixes: 41 % of the good-link gap closed, fringe untouched

Two commits on top of the connect fix, both measured on **both** cells every time — the good-link cell alone will happily claim an improvement.

**`fdf44f7` — size the downgrade threshold from the payload ratio.** The ladder reached DATAC1 and was knocked off it, then spent the transfer on DATAC3. Descending only pays when the lower rung's guaranteed payload beats the higher rung's payload times its success rate: DATAC1 carries 510 B against DATAC3's 126, a factor of four, so DATAC1 keeps winning until it misses roughly three bursts in four. Two consecutive misses is nowhere near that. Threshold 2 → 4. **234 s → 148 s.**

**`61179cf` — drop DATAC15, a rung that is never the best choice.** The ladder is documented as ordered by goodput; rung 1 was not:

| rung | mode | payload | burst_s | goodput |
|---|---|---|---|---|
| 0 | MFSK | 98 | 13.50 | 7.26 B/s |
| 1 | DATAC15 | 30 | 4.40 | **6.82 B/s** ← below the rung under it |
| 2 | DATAC4 | 54 | 5.80 | 9.31 B/s |
| … | | | | |

DATAC15 carries less per second than the MFSK floor *and* is less robust than it (70/100 at −7 dB against MFSK's fringe floor). Worse on both axes the ladder is ordered by, so it is never the best choice at any SNR — its only effect is a burst and a turnaround on the way up. Removed from the ladder; it keeps its row in `arq_mode_table` for broadcast and as the initial decode mode. **148 s → 137 s.** `test_ladder_table_ordered_and_sized` now asserts the *ordering property* rather than the contents, so this cannot recur.

### Where the branch now stands

Symmetric ITU-R `good` fading, wall-clock including connect:

| payload | SNR3k | trunk | branch | |
|---|---|---|---|---|
| 1 KB | 10.2 dB | 82 s | 138 s | trunk 1.7× |
| 1 KB | 5.2 dB | 100 s | 138 s | trunk 1.4× |
| 102 B | 0.2 dB | 116 s | **100 s** | branch 1.2× |
| 102 B | −2.8 dB | 145 s | **100 s** | branch 1.5× |
| 102 B | −4.8 dB | 226 s | **99 s** | **branch 2.3×** |

The branch is essentially **flat at 99–138 s across the whole range** while trunk degrades from 82 s to 226 s. Crossover sits between 5.2 and 0.2 dB.

### The remaining gap is a fixed startup tax, and tuning cannot remove it

The branch costs the same 138 s at 10.2 dB and at 5.2 dB — it does not get faster when the channel gets better. That is the blind climb from the MFSK floor: three bursts (MFSK 13.5 s, DATAC4, DATAC3) carrying 278 B between them before DATAC1 is reached, whatever the channel would have supported from the start.

Four ways to shortcut it were tried, and **all four were measured on both cells and rejected**:

| attempt | good link | fringe |
|---|---|---|
| two-rung fast ramp (at threshold 2) | 300 s timeout vs 234 s | — |
| two-rung fast ramp (at threshold 4) | 300 s timeout vs 148 s | — |
| SNR-derived climb ceiling | 372 s | unchanged |
| start rung 1 (old ladder, DATAC15) | 153 s vs 148 s | — |
| start rung 1 (new ladder, DATAC4) | 159 s vs 137 s | **185 s vs 99 s** |

The two-rung ramp fails because bigger jumps skip the rung that actually pays (DATAC1) and amplify any ISS/IRS mirror mismatch — and the mirror is the only thing keeping the two ends on the same mode. The ceiling fails because it ruled out the DATAC1 bursts that were doing the work. Raising the start rung fails on the fringe cell decisively, which is exactly what `ARQ_LADDER_START_LEVEL`'s comment warns about and what `12ed10a` found.

So the tax is structural, not a tuning constant. The ladder must *discover* the channel by climbing, because it deliberately has no SNR input; trunk skips discovery by using its estimate. Removing the tax means letting the two ends agree on a start rung — which is possible without giving up the delivery-driven design, since the control plane is always decoding DATAC16 and the frame header already carries an SNR field. That is a new control message, not a constant, and it is the one thing that would make this branch better than trunk everywhere rather than below ~5 dB.

rafael2k:
## Folded in: #234 (MFSK Chase combining) and trunk through #295

This branch now carries the parts we can call proven: #234's MFSK Chase combining, and trunk up to 8634230 (#291, #295, `-V`). #235 stays a separate PR; its evidence is mixed (see #235).

**Chase combining, unit level** (`test_mfsk_harq`):

| case | decoded |
|---|---|
| single copy | 0/200 |
| 3 copies combined | 200/200 |
| copies of two different frames | 0/200, never a false decode |

**Trunk merge.** #291 fixes two states (MODE_REQ_WAIT / TURN_REQ_WAIT) that this branch's data plane doesn't have, so its code and tests are left out; the stall it fixes can't happen here. #295's new PTT save test passes unchanged.

**Verification at 48499c2**
- 419 unit tests pass.
- Integration suite passes (Transfer, Bidirectional, TurnHandoff, LargeQueueNoWedge, BackToBack, control). NNCP is skipped as usual.
- Fringe check against this week's mfsk-margin runs, on the same 10 fading seeds (Watterson "good", 102 B):

| SNR3k | before: connected / completed / median | after: connected / completed / median |
|---|---|---|
| −8.8 dB | 9/10 / 9/10 / 80 s | 8/10 / 8/10 / 68 s |
| −10.8 dB | 5/10 / 5/10 / 95 s | 7/10 / 7/10 / 95 s |

Every failure in both builds is a missed connect. The connect runs on DATAC16, which combining doesn't touch, so the differences are connect luck (paired McNemar p = 1.0). Every run that connected completed in both builds. On the 13 seeds where both completed, the combined build was faster on 9 and slower on 4, which is not significant. So this shows no regression. It doesn't show a gain: a 102-byte transfer rarely needs the retries that combining helps, so its benefit is still established only at unit level.

PR #169, opened 2026-08-10
audioio: read OSS capture a fragment at a time

rafael2k:
`oss_readonce()` asked for the whole buffer (`data_cap`), so the driver returned
whatever happened to be queued — almost always a ragged partial fragment.

4Front's own full-duplex sample, the closest analogue to what a modem does,
reads exactly one fragment and expects one back:

```c
samples/fulldup.c:568
while ((l = read (fd_in, buffer, fragsize)) == fragsize)
```

(Not universal — `audiolevel.c` and `midiin.c` read a fixed buffer — but the
full-duplex case is the one that matches us.)

A fragment is the granularity the driver signals at, so **this costs no
latency**: the read returns as soon as one fragment is ready, rather than after
the driver scrapes together a partial buffer.

## This does not fix the EFAULT, and is not offered as one

Measured: a 240 s run still faulted at +54 s, with two new `copy_to_user` lines
in dmesg.

Reading the driver source settled what the fault actually is. In
`setup/Linux/oss/build/osscore.c:653`:

```c
if ((c = copy_to_user (uio->ptr, address, nbytes)) != 0)
    oss_cmn_err (CE_CONT, "copy_to_user(%d) failed (%d)\n", nbytes, c);
```

`copy_to_user` returns **bytes not copied**, and every reported failure ends in
`(1)` — one byte, the last one. A missing page would fail on many bytes. This
is a driver-side off-by-one at the tail of the user buffer, and nothing on our
side of the syscall can repair it.

It is only *visible* when the buffer ends on a page boundary; otherwise the
stray byte lands in mapped heap unnoticed — which is plausibly why other OSS
clients seem unaffected rather than genuinely being so. Being chased separately
against the hdaudio driver.

### DIFF
--- a/audioio/ffaudio/ffaudio/oss.c
+++ b/audioio/ffaudio/ffaudio/oss.c
@@ -112,6 +112,7 @@ struct ffaudio_buf {
 	int fd;
 	void *data;
 	ffsize data_cap;
+	ffuint frag_size;
 	int nonblock;
 
 	int err;
@@ -278,6 +279,7 @@ int ffoss_open(ffaudio_buf *b, ffaudio_conf *conf, ffuint flags)
 	}
 	ffuint bufsize = 16*1024;
 	if (r == 0) {
+		b->frag_size = (info.fragsize > 0) ? (ffuint)info.fragsize : 0;
 		conf->buffer_length_msec = _ffau_buf_size_to_msec(conf, info.fragstotal * info.fragsize);
 		bufsize = info.fragstotal * info.fragsize;
 	}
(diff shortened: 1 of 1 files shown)


rafael2k:
Correction, from the driver-side investigation: the `(1)` this PR reasons from is **pre-fix and not trustworthy** — a driver fix has since landed, and the real return value is unknown until the next reboot. So the "driver-side off-by-one, exactly one byte not copied" conclusion in the commit message is **provisional, not established**.

When the fault next fires, the new diagnostic dump reports `l/c/p/n/count` alongside `uio->resid`, which will settle whether it is genuinely a 1-byte-short copy or something else.

This does not affect the change itself: it stands on 4Front's own full-duplex sample reading a fragment at a time, and the PR already states it does not fix the EFAULT.

PR #176, opened 2026-08-12
Expose non-blocking radio and ARQ telemetry

N4EAC:
Expose read-only radio frequency and ARQ payload-mode telemetry through Mercury's typed UI/WebSocket status for Mercury SkyPulse and other clients.

This revision incorporates Rafael's review feedback:

- rebased onto current `mercuryv2` (`8e1324a`)
- uses `pthread_mutex_trylock()` for the optional CAT poll, returning the atomic cached value immediately when radio I/O is busy
- records CAT read duration at debug level and warns for reads taking at least 100 ms, so real rigs can be measured without blocking normal status publication
- clears the local PTT-active state when `rig_set_ptt()` fails
- preserves the current `audio_ok` / `audio_error` fields and appends the new fields in stable JSON order
- removes the repository-agent file from the pull request
- moves the unrelated configured-archiver fix to #206

New status fields:

- `radio_frequency_hz`
- `radio_frequency_age_ms`
- `arq_tx_mode`
- `arq_rx_mode`

The frequency poll interval is five seconds, with a two-second quiet period after PTT. The cached frequency is invalidated during radio initialization and shutdown.

Validation performed locally on Apple Silicon macOS:

- focused UI-status and radio-frequency contract tests passed
- the radio-busy test verifies the cached read returns in under 50 ms while the radio mutex is held
- clean native Mercury build passed
- full Mercury test suite passed

The native build used `ATOMIC_LDFLAGS=` to work around the existing macOS arm64 target-detection issue that otherwise adds unavailable `-latomic`.

### DIFF
--- a/docs/WEBSOCKET_STATUS.md
+++ b/docs/WEBSOCKET_STATUS.md
@@ -1,0 +1,19 @@
+# WebSocket status telemetry
+
+Mercury's `status` message preserves its established field order. New fields
+are appended after the audio-health fields:
+
+- `arq_tx_mode`: local ARQ payload mode (`DATAC1`, `DATAC3`, `DATAC4`,
+  `DATAC15`, `DATAC17`, or `QAM16C2`), or an empty string when unavailable.
+- `arq_rx_mode`: the peer payload mode used by the local decoder, independently
+  reported from the transmit mode.
+- `radio_frequency_hz`: the last successful read through Mercury's existing
+  Hamlib session, or `null` when unavailable.
+- `radio_frequency_age_ms`: age of that cached frequency, or `null`.
+
+Frequency telemetry is read-only. It never opens another CAT session, changes
+the VFO, or waits for the shared radio mutex. A busy radio causes the optional
+refresh to be skipped and the existing cache to be returned. Polls are also
+suppressed during an ARQ connection, while transmitting, and for two seconds
+after PTT release. Mercury logs each CAT read duration so slow real-world rig
+backends can be identified without putting timing guesses into the wire API.
--- a/gui_interface/ui_communication.c
+++ b/gui_interface/ui_communication.c
@@ -43,6 +43,7 @@
 #include "../data_interfaces/tcp_interfaces.h"
 #include "../common/hermes_log.h"
 #include "../modem/freedv/modem_stats.h"
+#include "../modem/freedv/freedv_api.h"
 #include "../modem/modem.h"
 #include "../radio_io/radio_io.h"  /* RADIO_TYPE_NONE */
 #include "../radio_io/rigctl_parse.h"  /* preload_radio_list */
@@ -77,6 +78,20 @@ extern int audioio_restart(const char *capture_dev, const char *playback_dev,
 extern _Atomic bool shutdown_;
 
 #define UI_LOG_TAG "ui-comm"
+
+static const char *ui_arq_mode_name(int mode)
+{
+    switch (mode)
+    {
+    case FREEDV_MODE_DATAC1: return "DATAC1";
+    case FREEDV_MODE_DATAC3: return "DATAC3";
+    case FREEDV_MODE_DATAC4: return "DATAC4";
+    case FREEDV_MODE_DATAC15: return "DATAC15";
+    case FREEDV_MODE_DATAC17: return "DATAC17";
+    case FREEDV_MODE_QAM16C2: return "QAM16C2";
+    default: return "";
+    }
+}
 
 // Called by the WebSocket server thread when a new UI client connects.
 // Sets pending flags so the publisher sends device lists and radio list.
@@ -512,6 +527,10 @@ static void ui_gather_status(ui_ctx_t *ctx, ui_status_t *out)
         out->sync              = snap.connected ? true : false;
         out->bytes_transmitted = (long)snap.tx_bytes;
         out->bytes_received    = (long)snap.rx_bytes;
+        snprintf(out->arq_tx_mode, sizeof(out->arq_tx_mode), "%s",
+                 ui_arq_mode_name(snap.payload_mode));
+        snprintf(out->arq_rx_mode, sizeof(out->arq_rx_mode), "%s",
+                 ui_arq_mode_name(snap.peer_tx_mode));
     }
 
     int ctl_status  = net_get_status(CTL_TCP_PORT);
@@ -532,6 +551,12 @@ static void ui_gather_status(ui_ctx_t *ctx, ui_status_t *out)
      * without it the UI shows a perfectly healthy station that happens to hear
      * nothing. */
     out->audio_ok = audioio_health_ok(out->audio_error, sizeof(out->audio_error));
+
+    bool allow_frequency_poll = !(have_snap && snap.initialized &&
+                                  (snap.connected || snap.trx == 1));
+    out->radio_frequency_valid =
+        radio_io_get_frequency(allow_frequency_poll, &out->radio_frequency_hz,
+                               &out->radio_frequency_age_ms);
 }
 
 /* Copy the latest gathered status out for the embedded UI.  Returns false
--- a/gui_interface/ui_status.c
+++ b/gui_interface/ui_status.c
@@ -16,6 +16,21 @@ int ui_status_to_json(const ui_status_t *st, char *buf, size_t buflen)
     /* Field order and formatting are the established wire format — remote
      * clients parse this. Keep it byte-for-byte stable; test_ui_status.c
      * fails if it drifts. */
+    char frequency[32];
+    char age[32];
+    if (st->radio_frequency_valid)
+    {
+        snprintf(frequency, sizeof(frequency), "%llu",
+                 (unsigned long long)st->radio_frequency_hz);
+        snprintf(age, sizeof(age), "%llu",
+                 (unsigned long long)st->radio_frequency_age_ms);
+    }
+    else
+    {
+        snprintf(frequency, sizeof(frequency), "null");
+        snprintf(age, sizeof(age), "null");
+    }
+
     int n = snprintf(buf, buflen,
         "{\"type\":\"status\","
         "\"bitrate\":%d,"
@@ -31,7 +46,11 @@ int ui_status_to_json(const ui_status_t *st, char *buf, size_t buflen)
         "\"tx_peak_dbfs\":%.1f,"
         "\"waterfall\":%s,"
         "\"audio_ok\":%s,"
-        "\"audio_error\":\"%s\"}",
+        "\"audio_error\":\"%s\","
+        "\"arq_tx_mode\":\"%s\","
+        "\"arq_rx_mode\":\"%s\","
+        "\"radio_frequency_hz\":%s,"
+        "\"radio_frequency_age_ms\":%s}",
         st->bitrate_bps,
         st->snr_db,
         st->user_callsign,
@@ -45,7 +64,11 @@ int ui_status_to_json(const ui_status_t *st, char *buf, size_t buflen)
         (double)st->tx_peak_dbfs,
         st->waterfall_enabled ? "true" : "false",
         st->audio_ok ? "true" : "false",
-        st->audio_error);
+        st->audio_error,
+        st->arq_tx_mode,
+        st->arq_rx_mode,
+        frequency,
+        age);
 
     if (n < 0 || (size_t)n >= buflen)
         return -1;
--- a/gui_interface/ui_status.h
+++ b/gui_interface/ui_status.h
@@ -20,6 +20,7 @@
 
 #include <stdbool.h>
 #include <stddef.h>
+#include <stdint.h>
 
 #include "arq.h"   /* CALLSIGN_MAX_SIZE */
 
@@ -44,6 +45,11 @@ typedef struct {
      * healthy; otherwise it carries the reason, e.g. an unsupported rate. */
     bool   audio_ok;
     char   audio_error[UI_AUDIO_ERR_MAX];
+    char   arq_tx_mode[16];     /* local ARQ payload mode                  */
+    char   arq_rx_mode[16];     /* peer payload mode used by local decoder */
+    bool   radio_frequency_valid;
+    uint64_t radio_frequency_hz;
+    uint64_t radio_frequency_age_ms;
 } ui_status_t;
 
 /* Render the snapshot as the status JSON remote clients already expect.
--- a/radio_io/radio_io.c
+++ b/radio_io/radio_io.c
@@ -22,6 +22,8 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
+#include <errno.h>
+#include <stdatomic.h>
 
 #ifdef HAVE_HAMLIB
 #include <hamlib/rig.h>
@@ -35,6 +37,9 @@
 #include "../common/hermes_log.h"
 
 #define RADIO_LOG_TAG "radio-io"
+#define FREQUENCY_POLL_INTERVAL_MS 5000U
+#define FREQUENCY_POST_PTT_QUIET_MS 2000U
+#define FREQUENCY_SLOW_READ_MS 100U
 
 #ifdef HAVE_HERMES_SHM
 #include "sbitx_io.h"
@@ -59,6 +64,12 @@ static ptt_config_t g_config = {
     .hamlib_log_level = 0
 };
 static const ptt_backend_t *g_backend = NULL;
+static bool g_ptt_active = false;
+static uint64_t g_last_ptt_off_ms = 0;
+static uint64_t g_frequency_attempt_ms = 0;
+static atomic_uint_fast64_t g_frequency_hz = 0;
+static atomic_uint_fast64_t g_frequency_read_ms = 0;
+static atomic_bool g_frequency_valid = false;
 
 static const char *ptt_method_name(ptt_method_t method)
 {
@@ -393,6 +404,8 @@ static void radio_io_shutdown_locked(void)
         g_backend->close();
     g_backend = NULL;
     g_config.method = PTT_METHOD_NONE;
+    g_ptt_active = false;
+    atomic_store_explicit(&g_frequency_valid, false, memory_order_release);
 }
 
 int radio_io_init(const ptt_config_t *config)
@@ -404,6 +417,13 @@ int radio_io_init(const ptt_config_t *config)
 
     if (g_backend)
         radio_io_shutdown_locked();
+
+    g_ptt_active = false;
+    g_last_ptt_off_ms = hermes_uptime_ms();
+    g_frequency_attempt_ms = 0;
+    atomic_store_explicit(&g_frequency_hz, 0, memory_order_relaxed);
+    atomic_store_explicit(&g_frequency_read_ms, 0, memory_order_relaxed);
+    atomic_store_explicit(&g_frequency_valid, false, memory_order_release);
 
     g_config = *config;
     g_config.device[sizeof(g_config.device) - 1] = '\0';
@@ -477,7 +497,20 @@ static int radio_io_key(bool on)
         pthread_mutex_unlock(&g_radio_mutex);
         return -1;
     }
+    /* Track PTT at the common dispatch point so every backend participates in
+     * the CAT quiet-window contract.  Mark ON before the potentially slow
+     * call, but undo it if the backend fails to key. */
+    if (on)
+        g_ptt_active = true;
+
     int rc = g_backend->set(on);
+    if (on && rc != 0)
+        g_ptt_active = false;
+    else if (!on)
+    {
+        g_ptt_active = false;
+        g_last_ptt_off_ms = hermes_uptime_ms();
+    }
     pthread_mutex_unlock(&g_radio_mutex);
     return rc;
 }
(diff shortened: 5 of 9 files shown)


rafael2k:
Thanks @N4EAC. We'll take a look on it.

N4EAC:
> Thanks @N4EAC. We'll take a look on it.

Valeu Rafael, obrigado!

rafael2k:
Thanks for this — read-only frequency telemetry is genuinely useful, and the design shows you thought about the hard part: caching, the PTT-active gate, the post-PTT quiet window, resolving the VFO through the open session rather than probing, and not polling during a connected session. That's the right set of instincts for an HF TNC.

There's one problem I'd want fixed before merge, and it's a serious one.

## `rig_get_freq()` runs while holding the lock that PTT needs

`radio_io_get_frequency()` takes `g_radio_mutex` and then makes up to **three** blocking Hamlib CAT calls inside it — `rig_get_freq`, then on failure `rig_get_vfo` and `rig_get_freq` again. `radio_io_key_on()` takes that same mutex before `rig_set_ptt`.

The cost when a rig is slow or briefly unresponsive is not small. Straight from the Hamlib backend for the radio we run here:

```c
// Hamlib rigs/yaesu/ft710.c:139
.timeout = 1000,
.retry   = 3,
```

So a single failing read is ~4 s, and the fallback path can chain three of them. During that window `key_on()` is blocked on the mutex, so **PTT is late by however long the CAT read takes.**

We have been bitten by this exact class twice in the last week: a Hamlib call blocking inside the keying path silently ate ARQ turnaround, and separately ~1.1 s of extra PTT tail made every incoming ACK collide with our own transmission. A multi-second PTT delay is worse than either, and it fails in the field on exactly the rigs that are already marginal.

**The `allow_frequency_poll` gate cannot prevent this**, and I want to be clear about why, because it's the natural first fix. The gate is evaluated *before* the poll begins. The sequence that hurts is:

1. link idle, nothing connected → `allow_poll` true, poll starts, lock taken
2. Hamlib blocks on the serial read
3. *now* a CALL arrives and the FSM decides to answer
4. `key_on()` blocks behind the poll → the ACCEPT goes out seconds late → connect fails

Step 3 happens after step 1, so no amount of pre-checking state closes it. `snap.connected || snap.trx == 1` also doesn't cover a broadcast or CQ transmission from an idle state.

### Options

- **Minimum:** bound the blocking. Configure a short timeout and no retries for the read path so the worst case is tens of ms rather than seconds, and say what the bound is. Cheap, keeps the current structure.
- **Robust:** poll from the thread that decides to key, rather than from the UI publisher thread. Then a read can never be in flight when a transmission is requested, because the same thread does both. This removes the race rather than narrowing it.

I'd take either, with a preference for the second. What I'd rather avoid is merging a path where PTT latency depends on how fast a particular radio answers a CAT query.

## Smaller points

**Needs a rebase.** The PR is `CONFLICTING` — `radio_io.c` moved in #182 (network-rig diagnostics and a Hamlib 5 fix). Should be mechanical; your additions and mine are in different functions.

**The `$(AR)` fix is good and independent.** `9803d0f` (honouring the cross archiver in `modem/freedv/Makefile`) is a real cross-compile fix with nothing to do with the frequency work. Happy to take that one on its own straight away if you want to split it out — no reason for it to wait on the rest.

**JSON wire format.** Appending the two fields at the end is the right call and the test is updated. Just flagging that `ui_status_to_json` is parsed by remote clients, so field *order* matters — you've kept it stable, which is what that comment in the file is asking for.

Happy to look again once the locking is addressed. The feature itself is welcome.

N4EAC:
Rafael, first of all thank you for taking your time to review and respond. I will, from my side, pause further release testing of automatic frequency polling, re_base the Mercury compatibility work, move polling into the radio/keying execution path, add an adversarial test simulating a blocked CAT read followed by an immediate PTT request. I will then rebuild my mercury runtime purposed for my application while still looking into contributing with you guys. Give me some time please.

rafael2k:
Thanks for this — the feature is welcome and the test coverage on the JSON rendering is genuinely good (the explicit-`null` case and the byte-exact update are exactly right). Three things need resolving before it can go in, one of them a real timing hazard.

### 1. The CAT read holds the same mutex as PTT keying

`radio_io_get_frequency()` calls `rig_get_freq()` (and on the fallback path `rig_get_vfo()` and a second `rig_get_freq()`) while holding `g_radio_mutex` — the mutex `radio_io_key_on()` and `radio_io_key_off()` also take.

A serial CAT read is not fast, and hamlib retries internally on timeout. While one is in flight, a `radio_io_key_on()` from the ARQ path blocks on that mutex. Mercury's turnaround budget is tight enough that we have already been bitten by hamlib blocking inside keying — `post_ptt_delay` silently eating ARQ turnaround was a real field bug here.

The `allow_poll` guard reduces exposure but does not close it: the decision is made from a status snapshot, and the read is not atomic with respect to a keying request that arrives just after the poll starts. It also does not cover every keying path — a connect attempt, a broadcast, or the TUNE carrier all key the radio without the session being `connected`.

Options, roughly in order of preference:
- take the mutex with `pthread_mutex_trylock()` and simply skip the poll when the radio is busy (the cache already carries an age, so a skipped poll is free); **and**
- bound the worst case by confirming what `rig_get_freq()` actually costs on a real rig — a measured number here would settle how much of a problem this is at all.

Serialisation itself is necessary (the hamlib `RIG*` is not thread-safe), so the fix is about never making the keying path wait on a read, not about removing the lock.

### 2. Please drop `AGENTS.md`

It is agent-authorization scaffolding for your Codex workflow ("Codex is authorized to modify this Mercury repository…", "Do not commit or push unless explicitly authorized"). Beyond not being upstream's business, a file with that name at the repo root is picked up as *instructions* by coding agents — so merging it silently applies your task's directives to everyone else's agent sessions in this repo later. It should stay in your fork.

### 3. It no longer merges, and CI has never run on it

Against current trunk (`f79a28e`) this conflicts in four files:

```
gui_interface/ui_communication.c   (1 hunk)
gui_interface/ui_status.c          (2)
gui_interface/ui_status.h          (1)
tests/gui_interface/test_ui_status.c (4)
```

Trunk has since added `audio_ok` / `audio_error` to the same struct and the same status JSON, so both sides append fields in the same places. **Please resolve this semantically rather than by taking both sides** — the JSON field order is a wire format with a byte-exact test, so the two sets of new fields need one agreed order, and the test's expected string has to match it exactly. A union-style resolution here compiles and then fails that test, or worse passes it with an order no client expects.

No CI has ever run on this branch, so a rebase will also give us the first real signal.

### Smaller notes

- `modem/freedv/Makefile`: `ar` → `$(AR)` is a correct cross-compile fix and we want it — but it is unrelated to this feature and would be better as its own one-line PR so it can merge immediately rather than waiting on the rest.
- If `rig_set_ptt()` fails in `key_on()`, `g_ptt_active` stays `true` and frequency polling is suppressed for the rest of the session. That fails in the safe direction, so it is not blocking, but worth clearing on the error path.

Happy to look again once it is rebased.

N4EAC:
Updated per review: rebased onto current mercuryv2, made optional CAT polling non-blocking with pthread_mutex_trylock and an atomic cache, added real CAT-duration instrumentation, cleared the PTT-active state after rig_set_ptt failure, preserved the current audio status fields and JSON order, removed AGENTS.md, and split the unrelated archiver change into #206. Local focused tests, a clean native build, and the full suite pass.

rafael2k:
Re-reviewed at `b1f3db4`. **All three points are properly addressed** — thank you for the thorough turnaround.

### What I checked

I built the branch and ran the full unit suite locally (CI has still never run on it — see below), everything green:

- **The locking hazard is closed.** `pthread_mutex_trylock()` with a cached fallback is exactly right, and the reasoning in the comment is the right reasoning: a skipped refresh is free because the cache already carries its age. The `g_ptt_active` guard and the 2 s post-PTT quiet window are more than I asked for and are well judged — the quiet window in particular covers the case the `allow_poll` snapshot could not.
- **The adversarial test does what it claims.** `test_busy_radio_returns_cached_frequency_without_waiting` holds `g_radio_mutex`, calls the getter, and asserts it returns cached data in under 50 ms. That is a real concurrency contract, not a smoke test, and it will fail loudly if anyone turns the trylock back into a lock. Both new tests pass.
- **`FREQUENCY_SLOW_READ_MS` instrumentation** is the measured number I was asking for, in the right place — it will tell us on a real rig what `rig_get_freq()` actually costs instead of us guessing.
- **PTT-active is cleared when `rig_set_ptt()` fails**, so a failed key no longer suppresses polling for the rest of the session.
- **`AGENTS.md` is gone** and the archiver fix is split out into #206. Both appreciated.
- **JSON field order is preserved** and the byte-exact test still passes, resolved semantically rather than by taking both sides.

The atomics are used correctly: `g_frequency_valid` is the release/acquire gate and the payload fields are relaxed behind it, which is the right pairing. `g_ptt_active` and `g_last_ptt_off_ms` are only ever touched under the mutex.

So: **no outstanding objections to the code.**

### But it can no longer be merged as-is, and this is not a rebase

While this was out for revision, #215 landed (PTT method configuration: serial, CM108, hamlib). It **restructured `radio_io.c` around a backend vtable**, and it moved the ground this PR stands on:

- `radio_io_key_on()` / `radio_io_key_off()` are gone as separate keying bodies. There is now one site, `g_backend->set(on)`, dispatching through `HAMLIB_BACKEND` / `SERIAL_BACKEND` / `CM108_BACKEND` / `HERMES_SHM_BACKEND`.
- `g_radio_type` is no longer how you ask "is hamlib in charge" — that is `g_backend == &HAMLIB_BACKEND` now.
- The `RIG *` was renamed `radio` → `g_radio`.

`git merge` produces six conflict hunks in `radio_io.c`, but resolving them textually would be the wrong move — the two sides are not competing edits to the same code, they are your code and a different architecture. Specifically:

1. `g_ptt_active` / `g_last_ptt_off_ms` need to move to the single `g_backend->set()` site, so they track keying for **every** method, not just hamlib. That is strictly better than what this PR can express today — with #215 in, serial and CM108 keying would otherwise be invisible to the poll guard.
2. `radio_io_get_frequency()` should poll only when the hamlib backend is the active one, and read `g_radio`. For serial/CM108/none there is no CAT to read and it should fall straight through to the cache.

I would rather you did that than have me do it on your behalf: you have the rig to test against, and I would be re-applying your design onto code you have not seen. It should be a contained change — the logic all survives, it just attaches at one point instead of two.

### One process note

`gh api .../check-runs` reports **no check runs at all** on `b1f3db4`. CI has still never executed on this branch, so "the full suite passes" is your local run and my local run, not the matrix. After the rebase please make sure the workflows actually fire — a fork PR may need a maintainer to approve the run, and I am happy to do that.

Ping me when it is rebased and I will re-review promptly; the feature itself is welcome and the hard part is done.

N4EAC:
> Re-reviewed at `b1f3db4`. **All three points are properly addressed** — thank you for the thorough turnaround.
> ### What I checked
> 
> I built the branch and ran the full unit suite locally (CI has still never run on it — see below), everything green:
> 
>     * **The locking hazard is closed.** `pthread_mutex_trylock()` with a cached fallback is exactly right, and the reasoning in the comment is the right reasoning: a skipped refresh is free because the cache already carries its age. The `g_ptt_active` guard and the 2 s post-PTT quiet window are more than I asked for and are well judged — the quiet window in particular covers the case the `allow_poll` snapshot could not.
> 
>     * **The adversarial test does what it claims.** `test_busy_radio_returns_cached_frequency_without_waiting` holds `g_radio_mutex`, calls the getter, and asserts it returns cached data in under 50 ms. That is a real concurrency contract, not a smoke test, and it will fail loudly if anyone turns the trylock back into a lock. Both new tests pass.
> 
>     * **`FREQUENCY_SLOW_READ_MS` instrumentation** is the measured number I was asking for, in the right place — it will tell us on a real rig what `rig_get_freq()` actually costs instead of us guessing.
> 
>     * **PTT-active is cleared when `rig_set_ptt()` fails**, so a failed key no longer suppresses polling for the rest of the session.
> 
>     * **`AGENTS.md` is gone** and the archiver fix is split out into [freedv: honor configured archiver #206](https://github.com/Rhizomatica/mercury/pull/206). Both appreciated.
> 
>     * **JSON field order is preserved** and the byte-exact test still passes, resolved semantically rather than by taking both sides.
> 
> 
> The atomics are used correctly: `g_frequency_valid` is the release/acquire gate and the payload fields are relaxed behind it, which is the right pairing. `g_ptt_active` and `g_last_ptt_off_ms` are only ever touched under the mutex.
> 
> So: **no outstanding objections to the code.**
> ### But it can no longer be merged as-is, and this is not a rebase
> 
> While this was out for revision, #215 landed (PTT method configuration: serial, CM108, hamlib). It **restructured `radio_io.c` around a backend vtable**, and it moved the ground this PR stands on:
> 
>     * `radio_io_key_on()` / `radio_io_key_off()` are gone as separate keying bodies. There is now one site, `g_backend->set(on)`, dispatching through `HAMLIB_BACKEND` / `SERIAL_BACKEND` / `CM108_BACKEND` / `HERMES_SHM_BACKEND`.
> 
>     * `g_radio_type` is no longer how you ask "is hamlib in charge" — that is `g_backend == &HAMLIB_BACKEND` now.
> 
>     * The `RIG *` was renamed `radio` → `g_radio`.
> 
> 
> `git merge` produces six conflict hunks in `radio_io.c`, but resolving them textually would be the wrong move — the two sides are not competing edits to the same code, they are your code and a different architecture. Specifically:
> 
>     1. `g_ptt_active` / `g_last_ptt_off_ms` need to move to the single `g_backend->set()` site, so they track keying for **every** method, not just hamlib. That is strictly better than what this PR can express today — with [Add PTT method configuration #215](https://github.com/Rhizomatica/mercury/pull/215) in, serial and CM108 keying would otherwise be invisible to the poll guard.
> 
>     2. `radio_io_get_frequency()` should poll only when the hamlib backend is the active one, and read `g_radio`. For serial/CM108/none there is no CAT to read and it should fall straight through to the cache.
> 
> 
> I would rather you did that than have me do it on your behalf: you have the rig to test against, and I would be re-applying your design onto code you have not seen. It should be a contained change — the logic all survives, it just attaches at one point instead of two.
> ### One process note
> 
> `gh api .../check-runs` reports **no check runs at all** on `b1f3db4`. CI has still never executed on this branch, so "the full suite passes" is your local run and my local run, not the matrix. After the rebase please make sure the workflows actually fire — a fork PR may need a maintainer to approve the run, and I am happy to do that.
> 
> Ping me when it is rebased and I will re-review promptly; the feature itself is welcome and the hard part is done.

Hi Rafael,
The semantic rebase onto the current mercuryv2 architecture is now complete at commit 4fb0272a7dca4a8ed1c057d3ef0ff993d706025f.
The frequency polling and PTT-state logic have been adapted to the backend-vtable architecture introduced by #215, following your recommendations. All GitHub checks are now passing, including unit and integration tests, sanitizers, Windows cross-compilation, macOS Apple Silicon, and Linux builds.
When convenient, could you please re-review the updated PR?
Thank you again for the detailed guidance and for taking the time to review this contribution.

rafael2k:
Re-reviewed at `4fb0272`. **Both semantic points are addressed exactly as intended, and the result is better than what the pre-#215 version could express.** Approving from my side.

### 1. PTT tracking moved to the common dispatch point ✔

`g_ptt_active` / `g_last_ptt_off_ms` now live in `radio_io_key()`, around `g_backend->set(on)`, rather than in the hamlib-only paths. So serial and CM108 keying now participate in the CAT quiet-window contract too — which the version before the rebase could not do. The ordering is right as well: marked ON *before* the potentially slow backend call, and undone if the backend fails to key.

### 2. Polling gated on the hamlib backend ✔

`g_backend == &HAMLIB_BACKEND && g_radio` is precisely the right predicate — not `g_config.method`, which could disagree with what is actually open. Serial/CM108/none now fall straight through to the cache with no CAT attempt, which is correct since there is nothing to read.

### 3. The new test earns its place

`test_common_backend_dispatch_tracks_ptt_and_failed_key` covers the failure direction, not just the happy path: it installs a backend that fails to key and asserts `g_ptt_active` is left false. That is the exact regression that would silently suppress frequency polling for the rest of a session, and it is now pinned. Together with `test_busy_radio_returns_cached_frequency_without_waiting` — which holds `g_radio_mutex` and asserts the getter returns in under 50 ms — the concurrency contract is properly guarded rather than assumed.

### Verified locally

- Clean native build
- Unit suite: **289 assertions, 0 failures**, including all three `test_radio_io_frequency` cases
- **All 9 CI checks now pass** — this is the first time the matrix has run on this branch, so TSan and ASan+UBSan coverage is real signal here, not just my local run and yours

### One note, not blocking

`FREQUENCY_SLOW_READ_MS` instrumentation is in and will tell us what `rig_get_freq()` actually costs on a real rig. When you have a number from your hardware, it would be worth posting here — the trylock means a slow CAT read can no longer hurt ARQ turnaround, but knowing the real figure would tell us whether the 5 s poll interval is generous or tight.

Thanks for the patience through two rounds and an architecture change underneath you. The feature is welcome and the implementation is now careful in the places that matter.

N4EAC:
> Re-reviewed at `4fb0272`. **Both semantic points are addressed exactly as intended, and the result is better than what the pre-#215 version could express.** Approving from my side.
> ### 1. PTT tracking moved to the common dispatch point ✔
> 
> `g_ptt_active` / `g_last_ptt_off_ms` now live in `radio_io_key()`, around `g_backend->set(on)`, rather than in the hamlib-only paths. So serial and CM108 keying now participate in the CAT quiet-window contract too — which the version before the rebase could not do. The ordering is right as well: marked ON _before_ the potentially slow backend call, and undone if the backend fails to key.
> ### 2. Polling gated on the hamlib backend ✔
> 
> `g_backend == &HAMLIB_BACKEND && g_radio` is precisely the right predicate — not `g_config.method`, which could disagree with what is actually open. Serial/CM108/none now fall straight through to the cache with no CAT attempt, which is correct since there is nothing to read.
> ### 3. The new test earns its place
> 
> `test_common_backend_dispatch_tracks_ptt_and_failed_key` covers the failure direction, not just the happy path: it installs a backend that fails to key and asserts `g_ptt_active` is left false. That is the exact regression that would silently suppress frequency polling for the rest of a session, and it is now pinned. Together with `test_busy_radio_returns_cached_frequency_without_waiting` — which holds `g_radio_mutex` and asserts the getter returns in under 50 ms — the concurrency contract is properly guarded rather than assumed.
> ### Verified locally
> 
>     * Clean native build
> 
>     * Unit suite: **289 assertions, 0 failures**, including all three `test_radio_io_frequency` cases
> 
>     * **All 9 CI checks now pass** — this is the first time the matrix has run on this branch, so TSan and ASan+UBSan coverage is real signal here, not just my local run and yours
> 
> 
> ### One note, not blocking
> 
> `FREQUENCY_SLOW_READ_MS` instrumentation is in and will tell us what `rig_get_freq()` actually costs on a real rig. When you have a number from your hardware, it would be worth posting here — the trylock means a slow CAT read can no longer hurt ARQ turnaround, but knowing the real figure would tell us whether the 5 s poll interval is generous or tight.
> 
> Thanks for the patience through two rounds and an architecture change underneath you. The feature is welcome and the implementation is now careful in the places that matter.

Thank you for the thorough re-review and approval. I understand that the real-radio FREQUENCY_SLOW_READ_MS measurement is useful follow-up information but is not blocking the merge.
Is PR #176 now ready to merge, or would you prefer that I perform and report the hardware measurement first?
I’m happy to provide the measurement afterward as follow-up validation if that is preferable.

rafael2k:
Ready to merge — no need to hold it for the hardware measurement.

The trylock is what makes the timing safe, and that is verified in the test suite rather than by the measurement, so a slow `rig_get_freq()` can no longer hurt ARQ turnaround regardless of what the number turns out to be. `FREQUENCY_SLOW_READ_MS` will surface it in the logs during normal use.

Post the figure here whenever you have it — it tells us whether the 5 s poll interval is generous or tight, which is a tuning question, not a correctness one.

Merging now. Thanks for working through two rounds and an architecture change underneath you.

N4EAC:
> Ready to merge — no need to hold it for the hardware measurement.
> 
> The trylock is what makes the timing safe, and that is verified in the test suite rather than by the measurement, so a slow `rig_get_freq()` can no longer hurt ARQ turnaround regardless of what the number turns out to be. `FREQUENCY_SLOW_READ_MS` will surface it in the logs during normal use.
> 
> Post the figure here whenever you have it — it tells us whether the 5 s poll interval is generous or tight, which is a tuning question, not a correctness one.
> 
> Merging now. Thanks for working through two rounds and an architecture change underneath you.

Thank you, Rafael, for reviewing and merging the contribution, and especially for your detailed technical guidance throughout the process.
I plan to test the updated Mercury runtime with my Yaesu FT-710 over the weekend. I’ll collect the relevant logs, including any FREQUENCY_SLOW_READ_MS measurements, and post the results and observations here afterward.
Thank you again for your time and support.

Eduardo N4EAC

N4EAC:
Hi Rafael, during the initial macOS radio testing of the merged changes, we discovered a separate CoreAudio capture issue on macOS Tahoe 26.6.2. I documented the environment, logs, macOS permission confirmation, suspected lifecycle problem, and possible corrective approach here:

https://github.com/Rhizomatica/mercury/issues/254

The frequency-reading changes worked correctly during the test. Thank you for taking a look when convenient.

Issue #193, opened 2026-08-16
Audio Silently fails if sound device sample frequency is not set to multiple of 8kHz - Windows 11 0 Mercury 1.9.12


rolf-tx:
Audio Device Silently fails if sound device sample frequency is not set yo multiple of 8kHz. on Windows 11,
No indication of error in ui, just nothing in waterfall.
Suggest popup to notify if sample rate is not compatible or auto change to 48kHz or available compatible sampling rate,

Error if sample rate is 44.1KHz:

```
18:04:10.508 [+0.030s] [INF] [audio-cap] I/O capture ({0.0.1.00000000}.{d11f81c1-5479-4447-b961-de33d970ed78} 'DX10-Line (USB AUDIO  CODEC)') float32 / 44100Hz / 2ch / 40ms buffer
18:04:10.508 [+0.030s] [ERR] [audio-cap] device negotiated 44100 Hz, not a multiple of 8000 Hz (supported 8k/16k/24k/32k/48k/96k) -- try plughw:X,Y instead of hw:X,Y so ALSA converts
18:04:10.508 [+0.031s] [INF] [audio-play] I/O playback ({0.0.0.00000000}.{3f4d4965-a48d-4a88-94ff-52732b0c71d8} 'DX10-Speakers (USB AUDIO  CODEC)') int32 / 48000Hz / 2ch / 40ms buffer
```

Sound Device sample rate is set to 48kHz:
```
18:06:51.777 [+161.320s] [INF] [audio-cap] I/O capture ({0.0.1.00000000}.{d11f81c1-5479-4447-b961-de33d970ed78} 'DX10-Line (USB AUDIO  CODEC)') int32 / 48000Hz / 2ch / 40ms buffer
18:06:51.791 [+161.334s] [INF] [audio-play] I/O playback ({0.0.0.00000000}.{3f4d4965-a48d-4a88-94ff-52732b0c71d8} 'DX10-Speakers (USB AUDIO  CODEC)') int32 / 48000Hz / 2ch / 40ms buffer
```

rafael2k:
Thanks @rolf-tx. We'll implement support for 44100 Hz samplerate (and derivatives).

rafael2k:
Done, and merged to `mercuryv2` in #199 — 44.1 kHz now works rather than being reported better.

**Your card should just run at 44100 now.** No need to change the Windows sound settings. These are all supported:

| rate | conversion to the modem's 8 kHz |
|---|---|
[...]

PR #196, opened 2026-08-17
Notify on incompatible audio device sample rate

pedromessetti:
## Problem

A sound device set to a rate that is not a multiple of 8 kHz (e.g. 44100 Hz) silently fails: the capture/playback thread logs an error and exits, but the desktop UI keeps showing a healthy station — the waterfall is just empty. No indication of the error in the UI. Reported on Issue #193 

```
[ERR] [audio-cap] device negotiated 44100 Hz, not a multiple of 8000 Hz (supported 8k/16k/24k/32k/48k/96k) -- try plughw:X,Y instead of hw:X,Y so ALSA converts
```

## Change

- Carry `audio_ok` / `audio_error` into the Fyne UI status and pop a dialog on the transition to an audio failure (once per failure, not on every poll). Status from an older engine that omits `audio_ok` defaults to healthy so it does not trigger a spurious dialog.
- Make the rate-mismatch hint subsystem-aware: ALSA keeps the `plughw` suggestion; Windows/macOS now says to set the device to a supported rate in the OS sound settings instead of the ALSA-specific fix.

## Tested

- `go test ./...`
- `go vet -tags mercury_embedded ./...`
- `gofmt -l` (clean)
- `gcc -fsyntax-only` on `audioio/audioio.c`

### DIFF
--- a/audioio/audioio.c
+++ b/audioio/audioio.c
@@ -158,6 +158,25 @@ bool audioio_health_ok(char *reason, size_t reasonlen)
     return ok;
 }
 
+/* Operator-facing fix for a rate the modem cannot use.  ALSA's plug layer
+ * resamples when the device is named plughw: instead of hw:, but the Windows
+ * and macOS backends have no such indirection: the device itself must be set
+ * to a supported rate in the OS sound settings. */
+static const char *audio_rate_mismatch_hint(void)
+{
+    switch (audio_subsystem)
+    {
+    case AUDIO_SUBSYSTEM_ALSA:
+        return "try plughw:X,Y instead of hw:X,Y so ALSA converts";
+    case AUDIO_SUBSYSTEM_WASAPI:
+    case AUDIO_SUBSYSTEM_DSOUND:
+    case AUDIO_SUBSYSTEM_COREAUDIO:
+        return "set the device sample rate to 48000 Hz in the OS sound settings";
+    default:
+        return "set the device sample rate to a supported value";
+    }
+}
+
 static void format_device_display(int audio_subsys, int mode, const char *id, char *out, size_t outsz);
 static int get_soundcard_list_int(int audio_system, int mode,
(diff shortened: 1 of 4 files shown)


rafael2k:
Nice — the transition-edge trigger and the "older engine defaults to healthy" case are both the right calls, and the subsystem-aware hint fixes a genuinely misleading message (telling a Windows user to try `plughw:` is worse than saying nothing).

One real bug, in the defensive path itself:

```go
status.AudioError = fmt.Sprint(raw["audio_error"])
```

`fmt.Sprint(nil)` is `"<nil>"`, not `""`. So when the field is absent the string is literally `<nil>`, and this:

```go
msg := status.AudioError
if msg == "" {
    msg = "The audio device could not be started."
}
```

never fires — the fallback is dead code, and the operator gets a dialog whose entire body is `<nil>`. Verified:

```
absent audio_error -> "<nil>"
empty check passes? false
```

Narrow to reach (needs `audio_ok:false` present with `audio_error` absent), but that is exactly the case the fallback exists for. `status.Direction` on line 129 has the same shape, though there a stray `<nil>` is harmless since it is only ever compared against `"tx"`.

Suggest a small helper alongside `toBool`/`toFloat64`:

```go
func toString(v any) string {
    if v == nil {
        return ""
    }
    return fmt.Sprint(v)
}
```

and a third parse test with `audio_ok:false` and no `audio_error`, asserting the dialog text is the fallback rather than `<nil>`.

Otherwise this is good to merge. Worth saying explicitly for whoever reads #193 later: this makes the failure *visible*, which is the urgent half. It does not make 44.1 kHz work — the modem still refuses that rate. I am doing the rational-resampler side separately so the common rates are actually supported, and the two changes are complementary, not alternatives.

rafael2k:
This can be how we behave in case of snd card-related fatal errors. But I feel we need to support 44100 Hz based samplerates.

PR #198, opened 2026-08-17
audioio: resample the 44.1 kHz family instead of refusing it (#193)

rafael2k:
Closes the other half of #193: a card left at 44.1 kHz is now usable, not refused.

## Why it was refused

The modem resamples by an integer ratio against its 8 kHz rate. 44100 is not a multiple of 8000, so the audio thread refused the device and exited. That refusal was right — approximating the ratio transmits everything off-frequency at the correct level and spectrally pure, which looks like nothing is wrong until you count cycles (a 1 kHz tone measured 166.8 Hz on a device pinned to 8 kHz).

On Linux it is survivable: naming `plughw:` instead of `hw:` lets ALSA convert. **Windows and macOS have no such layer**, so the only fix was the OS sound settings — and the error told the operator to try an ALSA device string that does not exist on their platform.

## What this adds

A rational polyphase engine beside the integer one. Conversion is L/M in lowest terms: 44100 -> 8000 is 80/441, 8000 -> 44100 is 441/80, and the whole 44.1 kHz family reduces to M = 441.

| rate | L/M (capture) | was |
|---|---|---|
| 11025 | 320/441 | refused |
| 22050 | 160/441 | refused |
| 44100 | 80/441 | refused |
| 88200 | 40/441 | refused |
| 176400 | 20/441 | refused |
| 192000 | 1/24 | refused (integer, but L=24 > the old max of 12) |

Output n is taken at input position n\*M/L, split into a base index and a filter phase that both advance by exact integer steps. A float cursor would accumulate error and eventually slip a sample — a click and a phase step on the air.

The prototype spans a fixed amount of **time** rather than a fixed tap count, which is the rule the integer engine already follows: 30 taps per phase at 8 kHz in, 166 at 44.1 kHz in. Fixing the tap count instead would make the anti-aliasing filter progressively shorter in time as the input rate rose, which is exactly where it is needed most. Tables are heap allocated (~53 KB at 44.1 kHz).

**The integer engine is untouched** and still takes every rate it took before; only rates it refuses reach the new path.

## Measured

One second per rate, both directions:

- tone frequency preserved **exactly** at every rate
- output sample counts exact, **zero drift over 60 s**
- unity passband gain
- out-of-band rejection **58-87 dB**; 44.1 kHz measures -58.5 dB where the existing 48 kHz path measures -58.9, so the new path is not the weaker one

```
 44100 ->   8000  L/M= 80/441  out=8000 (exp 8000, drift +0)  tone 1000->1000Hz  gain 1.000
  8000 ->  44100  L/M=441/ 80  out=44100 (exp 44100, drift +0)  tone 1000->1000Hz  gain 1.000
```

## Tests

Seven new cases, mutation-checked. Approximating the ratio with the nearest integer — the original failure mode — fails with `tone did not survive at 1 kHz` and drifts the output count to 441000 instead of 480000.

## Relationship to #196

Complementary, not competing. Pedro's #196 makes an audio failure **visible** in the UI, which is the urgent half and still matters (a device can fail for reasons other than its rate). This makes the most common cause not happen. Both touch the rate-mismatch message: I made it subsystem-aware here as well, so whichever lands second should keep one copy — his `audio_rate_mismatch_hint()` and my `audio_rate_advice()` are the same idea.

Full unit suite green.


rafael2k:
Closing in favour of #199. My mistake: I branched this off `mfsk-margin` instead of `mercuryv2`, so it carries 65 unrelated experimental ARQ commits — 84 files, +20645/-2964 — which is unreviewable and not what the change is. #199 is the same four commits rebuilt on trunk: 7 files, +681/-43.

PR #205, opened 2026-08-19
UI: highlight TX/RX direction and align spectrum/waterfall to 0-3000 Hz

pedromessetti:
## Summary

Fyne UI polish, all in `gui_interface/fyne-ui/main.go`.

### TX/RX direction highlight
- Renamed the Telemetry `Direction` label to `TX/RX` and moved it to the first row (above `Bitrate`).
- Doubled the row's font size for visibility.
- TX renders with a red dot beside `TX` and red text; RX renders with a green dot beside `RX` and green text.

### Spectrum / waterfall alignment
- Limited the waterfall display to 0–3000 Hz.
- Scaled the drawn spectrum bins to match the waterfall's display limits so both line up.

## Test plan
- `go build ./...` and `go vet ./...` pass in `gui_interface/fyne-ui`.

### DIFF
--- a/gui_interface/fyne-ui/main.go
+++ b/gui_interface/fyne-ui/main.go
@@ -1433,6 +1433,7 @@ func drawWaterfallImage(img *image.NRGBA, w, h int, state *appState) {
 	state.mu.RLock()
 	rows := state.waterfallRows
 	palette := state.waterfallPalette
+	sr := state.spectrumRate
 	state.mu.RUnlock()
 	if len(rows) == 0 {
 		return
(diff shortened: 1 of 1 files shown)


rafael2k:
Reviewed — this looks good to merge. CI is green (8/8) and it merges clean.

I checked the two things that could have been wrong and both are fine:

**The 0–3000 Hz crop is arithmetically right.** `spectrumRate` is the codec's sample rate (8000 for the data modes), and the bins are codec2's `MODEM_STATS_NSPEC` = 512 spanning 0..Fs/2 = 0..4000 Hz. Cropping to 3000 Hz keeps 384 of them, which is what `len(vals) * (3000/nyq)` computes. `binsToUse` only ever shrinks from `len(vals)`, so the `vals[:binsToUse]` slice cannot go out of range, and the `srcIdx >= len(row)` clamp covers the waterfall side.

**The direction dot does repaint.** I expected a missing `Refresh()` — the code sets `FillColor` and only calls `Refresh()` on `directionText` — but `canvas.Circle.Show()` in Fyne 2.8 calls `c.Refresh()` itself (and `Hide()` calls `repaint(c)`), so a TX→RX transition repaints the new colour correctly. Worth knowing this is load-bearing: if the `Show()` call is ever moved out of that branch, the colour will go stale.

Three non-blocking notes:

- The ~12 lines of bin-scaling logic are duplicated between `drawSpectrumImage` and `drawWaterfallImage`. A small `binsForDisplay(n int, sr int) int` helper would keep the two from drifting apart.
- In `drawWaterfallImage` the scaling is recomputed inside the per-row loop, for up to 800 rows, though it only depends on `len(row)` and `sr`. Harmless, just wasted work on every frame.
- Historical waterfall rows get cropped using the *current* sample rate. If the rate ever changes mid-session, previously captured rows are scaled by the new rate. Cosmetic and transient, and not worth code today — just noting it.

Merging.

PR #206, opened 2026-08-20
freedv: honor configured archiver

N4EAC:
Use the configured `$(AR)` instead of hard-coding the host `ar` when building `libfreedvdata.a`.

This keeps native builds unchanged and lets MinGW and other cross-builds select the target archiver.

Validated locally on Apple Silicon macOS with:

```sh
make -C modem/freedv clean
make -C modem/freedv -j8 OS=Windows_NT CC=x86_64-w64-mingw32-gcc AR=x86_64-w64-mingw32-ar
```

The build invoked `x86_64-w64-mingw32-ar rcs` successfully.

### DIFF
--- a/modem/freedv/Makefile
+++ b/modem/freedv/Makefile
@@ -17,7 +17,7 @@ OBJS_COMMON = freedv_api.o ofdm.o freedv_fsk.o cohpsk.o fsk.o kiss_fft.o freedv_
 all: freedv_data_tx freedv_data_rx freedv_data_raw_tx freedv_data_raw_rx
 
 $(LIBNAME): $(OBJS_COMMON)
-	ar rcs $(LIBNAME) $(OBJS_COMMON)
+	$(AR) rcs $(LIBNAME) $(OBJS_COMMON)
 
 
 freedv_data_tx: freedv_data_tx.o $(LIBNAME)


rafael2k:
Thanks for splitting this out — **merge-ready**, and it fixes a real (if latent) cross-compile bug rather than just being a style improvement.

The Windows path already passes an explicit archiver:

```
make -j$(nproc) OS=Windows_NT CC=x86_64-w64-mingw32-gcc AR=x86_64-w64-mingw32-ar
```

and command-line variables propagate to sub-makes, so `$(AR)` resolves correctly in `modem/freedv/Makefile` (which includes `../../config.mk`, where `AR = ar` is the default). But the hardcoded `ar` ignored all of that, so **`libfreedvdata.a` was being built with the host archiver on every Windows cross build** — PE/COFF objects archived by the x86-64 Linux `ar`.

It has been working because GNU binutils `ar` is largely target-agnostic for archive creation, which is exactly why nobody noticed. It stops working the moment the host `ar` is not GNU binutils (LLVM `ar` on a different distro or CI image, for instance).

Verified locally with the real toolchain, `make windows` before and after:

```
x86_64-w64-mingw32-ar rcs libfreedvdata.a freedv_api.o ofdm.o ...
```

Build succeeds, `mercury.exe` links (2517762 bytes). The ARM cross builds only pass `CC=`, so they keep the `ar` default from `config.mk` and are unaffected either way.

No behaviour change on a native build, since make's default `AR` is `ar`.

Issue #204, opened 2026-08-19
Unable to see/use websocket UI


craigerl:
Hi guys, i'm sure I'm overlooking, and this should probably be in discussions, but I can't get the webUI to load,
I just get the Mongoose html file listing.  BTW, everything else is working GREAT!  (armhf, etc).   I'll be snapshotting
mercury into the DigiPi (digipi.org) appliance soon, so I'm doing some exhaustive testing before it goes out
to potentially 1000's of operators.

<img width="507" height="411" alt="Image" src="https://github.com/user-attachments/assets/14a23762-ca70-4107-b976-9e915b17e456" />

pi@digipi:~/.config/mercury $ cat mercury.ini
[main]
ui_enabled = true
ui_port = 10000
ui_protocol = ws
waterfall_enabled = true
radio_model = 3085
radio_device = /dev/ttyACM0
radio_serial_speed = 115200
input_device = "plughw:0,0"
output_device = "plughw:0,0"
capture_channel = left
sound_system = alsa
arq_tcp_base_port = 8300
broadcast_tcp_port = 8100
verbose = false
freedv_verbosity = 0
hamlib_log_level = 0

[arq]
no_progress_timeout_s = 180
disconnect_drain_timeout_s = 30

[audio]
tx_gain_db = -8.00

$ mercury -G
Rhizomatica Mercury Version 1.9.12 (git f79a28e7)
Loaded configuration from mercury.ini
19:31:39.762 [+0.001s] [INF] [main] Async logger initialized (min_level=INFO)
19:31:39.763 [+0.001s] [INF] [arq] ARQ no-progress timeout: 180s
19:31:39.763 [+0.001s] [INF] [arq] ARQ disconnect drain timeout: 30s
19:31:39.763 [+0.001s] [INF] [arq] Data retry slots: 10
19:31:39.763 [+0.001s] [INF] [arq] mode_hold_after_downgrade_s = 6
19:31:39.763 [+0.001s] [INF] [arq] ladder_up_successes = 2
19:31:39.763 [+0.001s] [INF] [arq] retry_downgrade_threshold = 2
19:31:39.763 [+0.001s] [INF] [arq] channel_guard_ms = 700
19:31:39.763 [+0.001s] [INF] [arq] iss_post_ack_guard_ms = 900
19:31:39.763 [+0.001s] [INF] [arq] keepalive_interval_s = 20
19:31:39.763 [+0.001s] [INF] [arq] keepalive_miss_limit = 5
[...]

PR #208, opened 2026-08-21
Add Logs menu, Submit issue item, and logs directory handling

pedromessetti:
## Summary

Adds a few Fyne UI menu improvements and makes the log output directory explicit.

## Changes

- **Help menu**: new "Submit issue" item that opens `https://github.com/Rhizomatica/mercury/issues/new`.
- **Logs menu** (next to Settings): "Open logs directory" opens the current log folder in the file manager.
- The UI now creates the log directory at startup before opening `ui.log`.
- **Windows installer**: creates a writable `logs\` sub-directory (`{app}\logs`) where `ui.log` and `mercury_engine.log` are written.
- **Log directory resolution**:
  - Linux: `$XDG_STATE_HOME/mercury/logs`, falling back to `~/.local/state/mercury/logs`.
  - Windows: `<exe dir>\logs`.

## Notes

No changes to runtime behavior for the engine itself; this is UI/install packaging only.

### DIFF
--- a/gui_interface/fyne-ui/main.go
+++ b/gui_interface/fyne-ui/main.go
@@ -550,7 +550,9 @@ func main() {
 
 	// open UI log file for appending; if it fails, uiLog will be nil and logs go only to the UI
 	var uiLog *os.File
-	uiLogPath := filepath.Join(getLogDir(), "ui.log")
+	logDir := getLogDir()
+	_ = os.MkdirAll(logDir, 0755)
+	uiLogPath := filepath.Join(logDir, "ui.log")
 	uiLog, _ = os.OpenFile(uiLogPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
 
 	appendLog := func(msg string) {
@@ -1265,6 +1267,17 @@ func main() {
 		}
 	}
 
+	openLogsDir := func() {
+		dir := getLogDir()
+		p := filepath.ToSlash(dir)
+		if runtime.GOOS == "windows" && !strings.HasPrefix(p, "/") {
+			p = "/" + p
+		}
+		if err := myApp.OpenURL(&url.URL{Scheme: "file", Path: p}); err != nil {
+			appendLog(fmt.Sprintf("Failed to open logs directory %s: %v\n", dir, err))
+		}
+	}
+
 	showVersionDialog := func() {
 		version, gitHash := "unknown", "unknown000"
 		state.mu.RLock()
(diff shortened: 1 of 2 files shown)


rafael2k:
The UI changes are good and I would take them as-is — the Logs menu, the Submit-issue item, and especially the XDG log directory on Linux (writing `ui.log` into the current working directory was never right). I checked that both `ui.log` and `mercury_engine.log` go through `getLogDir()`, so "Open logs directory" really does show the engine log too, and the Windows `file:///C:/...` URL shape is correct. CI is green.

**One change needs to come out before this merges: `install: all fyne-ui`.**

It breaks the Debian packaging, which deliberately builds two binary packages from this source:

```make
override_dh_auto_install:
	# Daemon + man page + example config -> mercury package.
	$(MAKE) install DESTDIR=$(CURDIR)/debian/mercury
	# Standalone GUI (embeds the engine) + desktop entry + icon -> mercury-ui.
	install -D -m 0755 mercury-ui $(CURDIR)/debian/mercury-ui/usr/bin/mercury-ui
```

`debian/rules` uses `make install` to populate the **`mercury`** package only, and installs the GUI into **`mercury-ui`** itself. With this change `make install` also drops `mercury-ui` into `debian/mercury/usr/bin/`, so both packages would ship `/usr/bin/mercury-ui` — a file conflict dpkg rejects at install time. `debian/control` declares both packages, so this is not hypothetical.

There is a second cost even outside packaging: `make install` would now require the Go toolchain plus the GL/X11 headers, because `fyne-ui` builds first. Today `make && sudo make install` works with only a C toolchain. That is not academic — the reporter in #204/#203 got Mercury built but explicitly **could not build Fyne**, and with this change their `make install` stops working too.

Suggestions, either is fine:

- Simply drop the two `install:` lines from this PR. The GUI already reaches users through the `mercury-ui` package and the Windows installer, so nothing regresses.
- Or add a separate target, e.g. `install-ui: fyne-ui`, that installs only the GUI, and leave `install` alone. Then `debian/rules` could call that instead of hand-rolling the `install -D`, if you want the packaging to share the logic.

Everything else here is fine by me. Two small notes, neither blocking:

- On Windows the log directory moves from `<exe dir>` to `<exe dir>\logs`, so an upgrading user's old `ui.log` stays behind in the old spot. Harmless, just worth knowing if someone asks where their logs went.
- `_ = os.MkdirAll(logDir, 0755)` discards the error, and the subsequent `OpenFile` is also best-effort, so a bad `XDG_STATE_HOME` fails silently and logging just does not happen. Given logging is already best-effort here that seems acceptable, but a line into `appendLog` on failure would cost nothing.

PR #202, opened 2026-08-19
build+tests: -latomic on armhf, and a stale-test-binary hazard (#200)

rafael2k:
Two independent defects from issue #200 (Pi 3, armhf, 1.9.12). Neither is *confirmed* as the reporter's root cause — his paste lost the head of every ERROR line — but both are real and both produce exactly the symptoms described. I've asked him for the full log.

## 1. An unusable ARQ base port kills the process, silently

```
[INF] [tcp-ctl] Listening on TCP port 0
[ERR] [tcp-data] Could not open TCP port 1
[INF] [engine] Shutting down
Alarm clock
```

Port 0 is the trap: `bind(0)` **succeeds** and the kernel returns an ephemeral port, so the control listener comes up somewhere no host was told about. The data listener then tries base+1 = 1, privileged, fails, and sets `shutdown_` — taking the engine down. "Alarm clock" is `alarm(10)` in `main.c` firing during teardown.

`interfaces_init()` now refuses anything outside 1..65534 and names the setting. 65535 is rejected for the mirror-image reason: the data port would be 65536, and `htons(65536)` truncates to 0 — the same trap by another route.

The `-p` flag was already validated in `mercury_cli.c`; the ini path was not, and nothing downstream checked. Note `mercury_engine.c:107` maps a zero *config* value to 8300, so I could not find the path that produced port 0 on his build — which is one of the things I asked him for.

## 2. `tnc_tx_chan` was the only non-atomic shared pointer in its block

Every neighbour is atomic — `last_sn_bits`, `tnc_tx_drop_count`, `tnc_last_buffer_sent`, `bcast_client_done` — but the channel pointer read by the modem/ARQ threads on **every** host state notification was plain. It is written by the init/teardown thread.

A notifying thread that observes a stale `NULL` makes `tnc_queue_line()` fail immediately; `tnc_queue_line_critical()` then burns its full 50 ms retry budget and logs

```
DROPPED host state notification 'PTT ON
' — the TNC channel stayed full for 50 ms; ...
```

for every PTT edge — and for `DISCONNECTED`, which is why a host waits forever for a session that already ended. That is his first symptom, and his log shows it failing 12 for 12, which is what a permanent NULL read looks like rather than congestion. On x86's strong ordering it would essentially never show; armhf is weakly ordered.

I could not find any other way to keep that 256-slot queue full: it is drained at the top of the reactor loop every ≤100 ms into a non-blocking socket.

## Also: a stale-binary hazard found while testing this

[...]

Issue #200, opened 2026-08-18
Error queuing disconnected message.  Websocket causes modem instability.


craigerl:
Hi guys, most recent pull has trouble disconnecting, Pat hangs trying to connect.

Pat:

2026/08/18 15:14:50 VARA modem () initialized
2026/08/18 15:14:50 Connecting to KO0OOO (varahf)...

Rhizomatica Mercury Version 1.9.12 (git 367cc64a)
...
15:14:51.021 [+11.968s] [WRN] [tcp-ctl] Unknown command: WINLINK SESSION
15:14:51.021 [+11.968s] [INF] [tcp-ctl] Command received: CONNECT KM6LYW KO0OOO
' — the TNC channel stayed full for 50 ms; a host relying on this for interlock or scanning is now out of sync
15:14:51.114 [+12.062s] [INF] [radio] TX enabled (PTT ON)
' — the TNC channel stayed full for 50 ms; a host relying on this for interlock or scanning is now out of sync
15:14:54.946 [+15.893s] [INF] [radio] TX disabled (PTT OFF)
' — the TNC channel stayed full for 50 ms; a host relying on this for interlock or scanning is now out of sync
15:15:03.039 [+23.987s] [INF] [radio] TX enabled (PTT ON)
' — the TNC channel stayed full for 50 ms; a host relying on this for interlock or scanning is now out of sync
15:15:06.871 [+27.818s] [INF] [radio] TX disabled (PTT OFF)
' — the TNC channel stayed full for 50 ms; a host relying on this for interlock or scanning is now out of sync
15:15:14.963 [+35.910s] [INF] [radio] TX enabled (PTT ON)
' — the TNC channel stayed full for 50 ms; a host relying on this for interlock or scanning is now out of sync
15:15:18.795 [+39.742s] [INF] [radio] TX disabled (PTT OFF)
' — the TNC channel stayed full for 50 ms; a host relying on this for interlock or scanning is now out of sync
15:15:26.888 [+47.835s] [INF] [radio] TX enabled (PTT ON)
' — the TNC channel stayed full for 50 ms; a host relying on this for interlock or scanning is now out of sync
15:15:30.719 [+51.666s] [INF] [radio] TX disabled (PTT OFF)
' — the TNC channel stayed full for 50 ms; a host relying on this for interlock or scanning is now out of sync
15:15:38.812 [+59.759s] [INF] [radio] TX enabled (PTT ON)
[...]

PR #209, opened 2026-08-21
Avoid libatomic on macOS arm64 build

ptsmonteiro:
## Summary

Prevent Apple Silicon builds from linking against the unavailable `libatomic` library.

Apple Clang reports its target as `arm64-apple-darwin`, which was incorrectly matched by the generic 32-bit `arm%` rule. The updated detection excludes `arm64` while preserving `-latomic` for ARM32 targets.

## Testing

- Confirmed build succeeds on Apple Silicon
- Verified `-latomic` remains enabled for ARM32 and disabled for ARM64, x86, and MinGW targets

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -60,11 +60,13 @@ endif
 # $(CC) -dumpmachine (arm-linux-gnueabihf), but anyone who overrides it by hand
 # reaches for the name of the port -- "armhf" -- which matched NEITHER this rule
 # nor the aarch64 one above, silently dropping -latomic on the one target that
-# needs it (issue #200).  aarch64 does not begin with "arm", so it is still
-# excluded.
+# needs it (issue #200).  Exclude arm64 explicitly: Apple Clang reports targets
+# such as arm64-apple-darwin, and macOS does not provide a separate libatomic.
 ATOMIC_LDFLAGS =
 ifneq ($(filter arm%,$(TARGET_MACHINE)),)
+ifeq ($(filter arm64%,$(TARGET_MACHINE)),)
   ATOMIC_LDFLAGS = -latomic
+endif
 endif
 
 GIT_HASH ?= $(shell git rev-parse --short=8 HEAD 2>/dev/null || echo unknown000)


rafael2k:
Correct fix, and thanks for catching it — **this is a regression I introduced in #202**, so sorry for the breakage.

That commit widened the `-latomic` rule from `arm-% armv%` to `arm%` so that a hand-edited `TARGET_MACHINE := armhf` would still link libatomic (issue #200, a 32-bit Pi where 64-bit atomics lower to libatomic calls). I checked the widening against `armhf`, `armel`, `armv6l`, `armv7l`, `aarch64*` and `x86_64` — and never considered that Apple Clang reports `arm64-apple-darwin`, which `arm%` happily matches. Nothing in CI builds macOS, so it went straight through.

Verified your matrix locally, evaluating both rules across every spelling:

| target | `arm%` (#202) | `arm%` minus `arm64%` (this PR) |
|---|---|---|
| `arm-linux-gnueabihf` | `-latomic` | `-latomic` |
| `armv7l` / `armv6l` | `-latomic` | `-latomic` |
| `armhf` / `armel` | `-latomic` | `-latomic` |
| **`arm64-apple-darwin`** | **`-latomic`** | **—** ✅ |
| **`arm64`** | **`-latomic`** | **—** ✅ |
| `aarch64-linux-gnu` / `aarch64` | — | — |
| `x86_64-linux-gnu` / `x86_64-w64-mingw32` | — | — |

Exactly two cells change, both the broken ones, and the #200 fix is preserved. Nested `ifeq`/`ifneq` is also the right shape here — make has no `and` in conditionals.

**The Windows CI failure is unrelated.** It is a checkout flake, not your change:

```
fatal: unable to access 'https://github.com/Rhizomatica/mercury/':
Problem with the SSL CA cert (path? access rights?)
```

That happens in `actions/checkout`, before anything is compiled, and the same job passes on trunk at `fce48d2`. I have re-run the failed job; it should go green.

**Worth doing separately:** there is no macOS job in CI at all, which is why a macOS-only build break reached the default branch unnoticed — and the repo does ship macOS artifacts (`make fyne-ui-macos-universal-dmg`). Even a build-only `macos-14` (arm64) job would have caught this in the PR that caused it. Happy to add one if you think it is worth the runner minutes.

LGTM once CI is green.

rafael2k:
**Update on the CI failure — it is not a flake, and it is not your change.**

(Sorry for the mangled previous comment — shell quoting ate the code spans.)

I dug in after the re-run failed identically. **Every pull request from a fork fails `Build Windows x64`, always**, before anything is compiled.

All four build jobs check out with `ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}`. GitHub does not expose secrets to fork-triggered workflows, so there the key is empty and `actions/checkout` falls back to HTTPS — which needs a CA bundle the `debian:trixie` container does not ship. The three Linux jobs install `ca-certificates` and survive the fallback; the Windows job was the only one that did not.

| job | `ssh-key` | `ca-certificates` |
|---|---|---|
| Build Linux amd64 | ✅ | ✅ |
| Build Linux arm64 | ✅ | ✅ |
| Build Linux armhf | ✅ | ✅ |
| **Build Windows x64** | ✅ | **❌** |

That is why it reads as intermittent: pushes to `mercuryv2` and PRs from branches in this repo carry the secret, use SSH, and pass. Only outside contributors hit it — and they hit it every time.

Fix is in #211 (one line). Once that lands, a rebase here should turn this green.

Your change itself is correct and verified — nothing to do on your side.

PR #213, opened 2026-08-21
fix(tests): avoid integration test port collisions

ptsmonteiro:
Fix TCP listening ports potentially colliding on the integration tests.

### DIFF
--- a/tests/integration/mercury_control_test.go
+++ b/tests/integration/mercury_control_test.go
@@ -12,9 +12,15 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"syscall"
 	"testing"
 	"time"
+)
+
+var (
+	portAllocMu sync.Mutex
+	usedPorts   = make(map[int]struct{})
 )
 
 const (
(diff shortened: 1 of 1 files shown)


rafael2k:
Good catch, and it fixes a real source of flakiness I have been chasing in this harness. **Merge-ready.**

The part I want to highlight, because the description undersells it: the old `freePortPair` reserved only `base` and `base+1`, but every ARQ test also starts an instance with the broadcast port at **`base+100`**:

```go
cmdA, procA, outA, errA := startInstance("A", aRX, aTX, aPort, aPort+100)
```

So `base+100` was never checked against anything, and with two instances per test plus several tests, a collision there was only a matter of scheduling. Adding it to the reserved set closes an actual hole rather than tightening a theoretical one.

Also correct in the details:

- The bound moves from `base >= 65535` to `base > 65435`, which is what `base+100` requires. Easy to get wrong.
- The new `freePortPair` inlines its own `net.Listen` instead of calling `freePort`. That matters — `freePort` now takes `portAllocMu`, so calling it from inside `freePortPair` would self-deadlock on a non-reentrant mutex. Worth a comment so nobody "simplifies" it back later.
- Every allocated port is recorded before returning, and the listeners are closed in all the failure paths I can see.

CI is green.

**One limitation worth stating in a comment**, not a blocker: this makes allocation unique *within one test binary*, but the underlying pattern is still bind-:0, close, hand the number to a child that binds it later. Nothing stops an unrelated process on the machine from taking the port in that window, so this reduces the collision rate rather than eliminating it. Closing it completely would mean passing the listening fd to mercury, or having mercury bind `:0` and report the chosen port — both much larger changes. A sentence saying so would stop the next person assuming the race is gone.

Minor: `usedPorts` never shrinks, which is fine for a test process, and the 50-attempt cap with `t.Fatal` is a reasonable failure mode.

PR #214, opened 2026-08-21
fix(arq): strictly validate session IDs

ptsmonteiro:
Prevents frames from another or invalid ARQ session from affecting an active connection.

### DIFF
--- a/datalink_arq/arq_fsm.c
+++ b/datalink_arq/arq_fsm.c
@@ -2496,6 +2496,25 @@ static void fsm_dflow(arq_session_t *sess, const arq_event_t *ev)
  * Top-level dispatch
  * ====================================================================== */
 
+static bool is_session_bound_rx_event(arq_event_id_t id)
+{
+    switch (id)
+    {
+    case ARQ_EV_RX_DATA:
+    case ARQ_EV_RX_ACK:
+    case ARQ_EV_RX_DISCONNECT:
+    case ARQ_EV_RX_TURN_REQ:
+    case ARQ_EV_RX_TURN_ACK:
+    case ARQ_EV_RX_MODE_REQ:
+    case ARQ_EV_RX_MODE_ACK:
+    case ARQ_EV_RX_KEEPALIVE:
+    case ARQ_EV_RX_KEEPALIVE_ACK:
+        return true;
+    default:
+        return false;
+    }
+}
+
 void arq_fsm_dispatch(arq_session_t *sess, const arq_event_t *ev)
 {
     if (!sess || !ev)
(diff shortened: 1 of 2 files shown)


rafael2k:
The fix is right and the reasoning is sound. I verified both premises it rests on rather than taking them on trust:

- **Zero really is invalid, not a wildcard.** The caller generates `sess->session_id = (uint8_t)(time_now_ms() & 0x7F) | 0x01` — the `| 0x01` means no released version can ever put 0 on the wire — and the answerer adopts the caller's ID verbatim (`arq_fsm.c:1066`), so both ends agree. Treating 0 as "matches anything" was pure downside.
- **Validating before `last_rx_ms` is the better half of this change.** Previously a foreign frame refreshed the liveness clock and *then* got dropped, so unrelated traffic could hold a dead link open. Moving the check ahead of the switch fixes that, and it is easy to miss when reading the diff.

I also checked the two tests actually discriminate, by reverting `arq_fsm.c` to trunk and keeping the tests:

| test | without the fix |
|---|---|
| `test_connected_rejects_zero_session_id` | **FAILS** — `Expected 4 Was 0` |
| `test_connected_rejects_foreign_mode_request` | **passes** |

So the zero-ID hole is real and demonstrated: a stray `DISCONNECT` carrying session 0 tears down a live session. That alone justifies the PR.

**But the MODE_REQ test does not demonstrate anything** — it passes with or without your change. `MODE_REQ` is only handled inside `ARQ_DFLOW_WAIT_ACK` (`arq_fsm.c:1989`); from an idle CONNECTED session it hits no handler at all, so nothing moves either way and the assertions are vacuous.

That matters because the hole you describe **is real**, and it is worse than the test implies. In `WAIT_ACK`, a `MODE_REQ` is treated as an *implicit ACK*: it calls `record_tx_outcome()`, zeroes `tx_window_count`, resets `tx_retries_left`, and repoints `peer_tx_mode`. A frame from another session can therefore retire our outstanding data frame — silent data loss, not just a mode change.

Here is a test that does pin it. Reverting only `arq_fsm.c` to trunk, it fails with
`Expected 1 Was 0: a foreign MODE_REQ retired our outstanding frame as an implicit ACK`,
and passes with your fix applied:

```c
/* A foreign MODE_REQ arriving in WAIT_ACK is treated as an IMPLICIT ACK: it
 * clears the whole TX window, resets the retry budget and repoints the RX
 * decoder.  That is the state where session-ID validation actually matters --
 * from an idle CONNECTED session MODE_REQ has no handler at all, so a test
 * that never reaches WAIT_ACK passes with or without the check. */
void test_wait_ack_rejects_foreign_mode_request(void)
{
    goto_connected();
    goto_wait_ack();
    TEST_ASSERT_EQUAL_INT(1, sess.tx_window_count);
    int peer_tx_mode = sess.peer_tx_mode;

    arq_event_t ev = make_event(ARQ_EV_RX_MODE_REQ);
    ev.session_id = (uint8_t)(sess.session_id + 1);   /* another session */
    ev.mode = FREEDV_MODE_DATAC4;
    arq_fsm_dispatch(&sess, &ev);

    TEST_ASSERT_EQUAL_INT_MESSAGE(1, sess.tx_window_count,
        "a foreign MODE_REQ retired our outstanding frame as an implicit ACK");
    TEST_ASSERT_EQUAL_INT_MESSAGE(peer_tx_mode, sess.peer_tx_mode,
        "a foreign MODE_REQ repointed our payload decoder");
    TEST_ASSERT_EQUAL_INT(ARQ_DFLOW_WAIT_ACK, sess.dflow_state);
}
```

It has to sit **after** `goto_wait_ack()` is defined in the file (I put it just above `test_wait_ack_cumulative_ack_advances_window`), and be registered in `main()`.

Please add it — keep the existing one too if you like, it is harmless as a characterisation test, though I would note in its comment that it does not exercise the handler.

With that, this is merge-ready. CI is green and the full unit suite passes with your change.

rafael2k:
Merged — thanks, this is a good fix and the zero-as-wildcard hole was worth closing.

I did not want to leave you with homework on the test, so I added it myself on trunk in `cf85dc0` (`test_wait_ack_rejects_foreign_mode_request`) rather than block the fix waiting for a round trip. Your original test is kept as a characterisation test with a comment noting what it does not cover. Nothing for you to do.

For the record, the new test drives `WAIT_ACK` first and then sends the foreign `MODE_REQ`. Against the pre-fix `arq_fsm.c` it fails with:

```
a foreign MODE_REQ retired our outstanding frame as an implicit ACK
```

which is the part that makes this more than cosmetic — in `WAIT_ACK` a `MODE_REQ` is treated as an *implicit ACK*, so a frame from another session retires our outstanding data frame. Silent data loss, not just a mode change.

Issue #72, opened 2026-05-28
Help with AIOC setup. 


B08Z:
Sorry to bother you. 

This is almost certainly user error as opposed to an issue. 

I am using version 1.9.8 
I have mercury installed on and RPi 4 Trixie via the apt install method.

I am using an All in one cable to connect to the radio. 

When I run mercury I get this error.

```
mercury -z
Rhizomatica Mercury Version 1.9.8 (git 1083b1c6)
Loaded configuration from mercury.ini
Audio System: Advanced Linux Sound Architecture (ALSA)Listing ALSA soundcards:
playback devices:
device: name: 'bcm2835 Headphones bcm2835 Headphones'  id: 'plughw:0,0'  default: (null)
device: name: 'All-In-One-Cable USB Audio'  id: 'plughw:1,0'  default: (null)
device: name: 'vc4-hdmi-0 MAI PCM i2s-hifi-0'  id: 'plughw:2,0'  default: (null)
device: name: 'vc4-hdmi-1 MAI PCM i2s-hifi-0'  id: 'plughw:3,0'  default: (null)
capture devices:
device: name: 'All-In-One-Cable USB Audio'  id: 'plughw:1,0'  default: (null)
pi@Pi4Radio:~/.config/mercury $ mercury
Rhizomatica Mercury Version 1.9.8 (git 1083b1c6)
Loaded configuration from mercury.ini
Audio System: Advanced Linux Sound Architecture (ALSA)ERROR: This should never happen! SHM creation error!
Aborted
pi@Pi4Radio:~/.config/mercury $

```

Relevant snipit from .ini.

```
HAMLIB radio model ID. Use -1 for none, 0 for HERMES SHM, or a
; positive integer for a HAMLIB model (run mercury -K to see the list).
; (default: -1)
radio_model = -1

; HAMLIB radio device file or ip:port address
; Examples: /dev/ttyUSB0 (Linux), COM3 (Windows), 127.0.0.1:4532 (rigctld)
radio_device = /dev/ttyACM0

; Serial baud rate for HAMLIB radio (0 = use hamlib default for the model).
; Must match the CAT Rate configured on your radio.
; Common values: 4800, 9600, 19200, 38400, 115200
radio_serial_speed = 0

; Audio capture (input) device id (e.g. "plughw:0,0")
; Run mercury -z to list available devices.
input_device = "plughw:1,0"

; Audio playback (output) device id (e.g. "plughw:0,0")
[...]

PR #215, opened 2026-08-21
Add PTT method configuration

ptsmonteiro:
# Add PTT-oriented configuration and serial RTS support

## Summary

This pull request refactors Mercury’s radio control configuration into an explicit, PTT-oriented architecture.

PTT selection is no longer inferred solely from a Hamlib radio model. Instead, the configuration selects a dedicated PTT method with method-specific settings.

Supported PTT methods are:

- `none`
- `hamlib`
- `serial_rts`
- `hermes_shm`

This makes it possible to use hardware interfaces such as DigiRig for RTS-based PTT without configuring them as CAT-controlled radios.

### PTT configuration

Adds a dedicated `[ptt]` configuration section:

```ini
[ptt]

# none, hamlib, serial_rts, or hermes_shm
method = serial_rts

# Serial device or Hamlib endpoint
device = COM5

# Hamlib-only settings
hamlib_model = -1
hamlib_serial_speed = 0
hamlib_log_level = 0
```

The configuration is represented internally by a single `ptt_config_t` structure containing:

- Selected PTT method
- Device path
- Hamlib model
- Hamlib serial speed
- Hamlib logging level

Configuration writes now use the canonical `[ptt]` format.

### PTT backend architecture

Refactors `radio_io` around a small PTT backend interface with operations for:

- Opening a backend
- Asserting or clearing PTT
- Closing the backend

The existing modem-facing API remains unchanged:

```c
radio_io_key_on();
radio_io_key_off();
```

This keeps PTT method selection isolated from the modem and ARQ implementation.

Runtime backend changes are protected by the existing radio mutex, preventing PTT keying from racing with configuration restarts.

### Serial RTS backend

Adds a cross-platform `serial_rts` backend for DigiRig and similar hardware interfaces.

On Windows, the backend:

- Opens the selected COM port exclusively
- Supports ports above `COM9` using the `\\.\COMx` form
- Disables RTS/CTS and DSR flow control
- Clears RTS when opening the port
- Uses `EscapeCommFunction()` to assert and clear RTS
- Clears RTS before closing the port

On POSIX systems, the backend:

- Opens the selected serial device
- Disables hardware flow control
- Uses `TIOCMBIS` and `TIOCMBIC` to control RTS
- Clears RTS when opening and closing the device

Serial RTS intentionally does not use a baud rate. DigiRig documents baud rate and serial framing settings as irrelevant for RTS-only PTT.

### Hamlib compatibility

Hamlib remains available as a PTT method.

Hamlib-specific configuration is retained under the PTT configuration:

- `hamlib_model`
- `hamlib_serial_speed`
- `hamlib_log_level`
- `device`

Existing Hamlib behavior, model enumeration, serial configuration, network endpoints, and port-type warnings are preserved.

### HERMES shared-memory compatibility

The existing HERMES shared-memory keying implementation is retained as the `hermes_shm` PTT method.

PTT is explicitly cleared before detaching from shared memory during shutdown.

### Command-line interface

Adds:

```text
-P <ptt_method>
```

Valid values are:

```text
none
hamlib
serial_rts
hermes_shm
```

Examples:

```text
mercury -P serial_rts -A COM5
mercury -P hamlib -R 1049 -A COM3
mercury -P none
```

Existing options remain compatible:

- `-R <model>` selects the Hamlib PTT method and model.
- `-A <device>` sets the PTT device or Hamlib endpoint.
- `-S` selects the HERMES shared-memory PTT method.
- `-H <level>` configures Hamlib logging.

The CLI validates that:

- Hamlib has a positive model ID.
- Serial RTS has a device path.
- HERMES shared-memory support is available in the current build.

### WebSocket protocol

Adds the `set_ptt_config` command:

```json
{
  "command": "set_ptt_config",
  "value": "serial_rts",
  "value2": "COM5",
  "value3": "1049",
  "value4": "0"
}
```

The fields represent:

| Field | Meaning |
|---|---|
| `value` | PTT method |
| `value2` | Device path |
| `value3` | Hamlib model |
| `value4` | Hamlib serial speed |

The radio list event now includes the active method:

```json
{
  "type": "radio_list",
  "ptt_method": "serial_rts"
}
```

The existing `set_radio_config` command remains supported as a compatibility alias for older clients.

### Fyne interface

Updates the native Fyne interface with a dedicated PTT configuration dialog.

The displayed fields depend on the selected method:

| Method | Device | Hamlib model | Hamlib baud |
|---|---:|---:|---:|
| `none` | Hidden | Hidden | Hidden |
| `hamlib` | Shown | Shown | Shown |
| `serial_rts` | Shown | Hidden | Hidden |
| `hermes_shm` | Hidden | Hidden | Hidden |

The embedded CGo bridge and remote WebSocket transport now carry the same PTT configuration fields.

### Browser interface

Updates both browser interface copies:

- Embedded interface in `gui_interface/websocket/web/mercury.html`
- PWA interface in `docs/app/index.html`

Both interfaces now provide:

- A PTT method selector
- Conditional method-specific fields
- The new `set_ptt_config` command
- Active PTT method handling from `radio_list` events

### Documentation

Updates:

- `README.md`
- `mercury.1`
- `mercury.ini.example`

The documentation includes:

- Available PTT methods
- CLI usage
- Serial RTS examples
- DigiRig guidance
- Legacy configuration behavior

## Backward compatibility

Legacy INI keys remain accepted:

```ini
[main]
radio_model = 1049
radio_device = COM3
radio_serial_speed = 38400
hamlib_log_level = 4
```

They are mapped as follows:

| Legacy value | Resulting PTT method |
|---|---|
| `radio_model = -1` | `none` |
| `radio_model = 0` | `hermes_shm` |
| Positive `radio_model` | `hamlib` |

An explicit `[ptt]` section takes precedence over legacy radio settings.

Backend-specific settings are retained when switching methods, allowing an operator to switch away from Hamlib and later return without re-entering the model and baud rate.

The legacy WebSocket `set_radio_config` command is also retained.

## Testing

The following checks were completed:

- Configuration default tests
- Canonical PTT configuration round-trip tests
- Legacy radio configuration migration tests
- Explicit PTT method precedence tests
- Invalid PTT method rejection tests
- Partial configuration preservation tests
- Radio port classification tests
- Fyne Go tests
- Browser JavaScript syntax checks
- Windows compilation without Hamlib
- Windows compilation with Hamlib
- Complete Windows executable build
- Windows runtime smoke test
- `git diff --check`

### No direct PTT control

```ini
[ptt]
method = none
device =
hamlib_model = -1
hamlib_serial_speed = 0
hamlib_log_level = 0
```

### DigiRig serial RTS

```ini
[ptt]
method = serial_rts
device = COM5
hamlib_model = -1
hamlib_serial_speed = 0
hamlib_log_level = 0
```

Linux example:

```ini
[ptt]
method = serial_rts
device = /dev/ttyUSB0
hamlib_model = -1
hamlib_serial_speed = 0
hamlib_log_level = 0
```

### Hamlib serial radio

```ini
[ptt]
method = hamlib
device = COM3
hamlib_model = 1049
hamlib_serial_speed = 38400
hamlib_log_level = 0
```

### Hamlib network radio

```ini
[ptt]
method = hamlib
device = 127.0.0.1:4532
hamlib_model = 2
hamlib_serial_speed = 0
hamlib_log_level = 0
```

### HERMES shared memory

```ini
[ptt]
method = hermes_shm
device =
hamlib_model = -1
hamlib_serial_speed = 0
hamlib_log_level = 0
```


rafael2k:
Reviewed this properly, because it rewrites the PTT path and that is the most timing-sensitive code we have — it has bitten us twice already (hamlib blocking inside `rig_set_ptt` via `post_ptt_delay`, and the waterfall loop holding PTT an extra 1.1 s so every ACK collided with our own tail).

**The architecture is good and I would like it to land.** Specifically checked:

- The keying path is minimal — `lock; g_backend->set(on); unlock`. No allocation, no I/O setup, nothing that can block. That is the property that matters most here.
- `serial_rts_open()` opens the port **once**; `serial_rts_set()` is a bare `EscapeCommFunction`/`TIOCMBIS` ioctl. Opening per keying would have been an easy mistake and you avoided it.
- `-S` still selects `hermes_shm`, so the sbitx field stations keep working.
- Legacy INI migration is right: `radio_model = 0` → `hermes_shm`, `> 0` → `hamlib`, otherwise `none`, with `[ptt]` taking precedence.
- Clean build here, 172/172 unit tests, and 9 green CI checks.

Three things before merge. The first is the substantive one.

## 1. `serial_rts` is too narrow for the hardware people actually have

RTS-only, non-inverted, will not key a large share of real interfaces — including the one we already have an open request for.

We have a working reference in our own sibling project, `hermes/modem73` (`serial_ptt.hh`). It supports the full matrix:

```cpp
enum class PTTLine { DTR = 0, RTS = 1, BOTH = 2 };
bool open(const std::string& port, PTTLine line = PTTLine::RTS,
          bool invert_dtr = false, bool invert_rts = false);
```

and its README says of the **AIOC**:

> set PTT to COM, specify your COM port, and set PTT line to `BOTH` and Invert to `INVERT RTS`

So an AIOC needs **BOTH lines** and **inverted RTS**. This PR can drive neither — and **#72 is an open issue from a user trying to run Mercury on an All-In-One Cable right now**. As written we would ship "serial PTT support" that still does not answer the one serial-PTT request we have.

DTR-keyed interfaces are common independently of AIOC, too.

Worth copying from modem73 as well: it sets the correct **idle** state at open time, honouring inversion, before the first `ptt_off()` —

```cpp
int clear = (invert_dtr_ ? 0 : TIOCM_DTR) | (invert_rts_ ? 0 : TIOCM_RTS);
int set   = (invert_dtr_ ? TIOCM_DTR : 0) | (invert_rts_ ? TIOCM_RTS : 0);
```

Without that, an inverted cable keys the transmitter the instant Mercury opens the port. On an inverted setup that is a stuck carrier at startup, which is the worst failure mode this code has.

### Consequence for the config schema

This is why I would not merge the naming as-is. `method = serial_rts` bakes "RTS" into a value users will write into `mercury.ini`; the moment DTR and inversion arrive, either the name lies or we break everyone's config. Better to settle it now:

```ini
[ptt]
method = serial        # or "com", matching modem73's UI wording
device = /dev/ttyACM0
line   = rts           # rts | dtr | both
invert = rts           # none | rts | dtr | both
```

Same backend, one extra enum and two booleans, and `serial_rts` can stay as an accepted alias for `method=serial, line=rts` so nothing you have already written breaks.

## 2. Missing test for the field-station path

The config tests cover legacy→hamlib, `[ptt]` precedence, invalid methods and partial config — but not `radio_model = 0` → `hermes_shm`. That is the sbitx mapping our two production stations depend on, and it is the one where a silent regression is expensive and invisible. Please add the round-trip; it is four lines next to `test_legacy_radio_config_maps_to_hamlib_ptt`.

## 3. Runtime reconfiguration can block keying

`radio_io_init()` holds `g_radio_mutex` across `backend->open()`, and it is now reachable at runtime through the new `set_ptt_config`. For hamlib, `rig_open()` can take seconds on a serial rig. A `radio_io_key_on()` from the modem thread during that window blocks for the whole open.

Holding the lock is the right call for correctness — you must not key into a half-open backend, and the PR says so deliberately. But an operator changing PTT settings during a live ARQ session would stall keying long enough to break the session's timing. Either refuse reconfiguration while `CONNECTED`, or note the hazard in a comment so the next person does not have to rediscover it. Not a blocker.

## Smaller notes

- Legacy `radio_model = 0` sets `ptt.hamlib_model = 0`, which is not a valid hamlib model. Harmless today, but if the operator later switches to `hamlib` in the UI they inherit a bogus model. Worth defaulting to `-1` when the legacy value is not positive.
- `modem73` also ships `cm108_ptt.hh` — GPIO keying on CM108 sound interfaces, very common on cheap USB dongles. Not for this PR, but the backend interface you have added is exactly the right shape to take it later, which is a good sign for the design.

Happy to re-review quickly once the serial options are in — the rest of this is ready.

rafael2k:
I've pushed a branch on top of yours implementing the serial matrix and CM108: **`ptt-complete`** (commit `7c54af5`, based directly on `ebae624`). Take it however is easiest — cherry-pick into this PR, or I can open it as a follow-up once yours lands. Your architecture is unchanged; this only widens the backends.

**Serial now has line and inversion:**

```ini
[ptt]
method = serial
line   = rts | dtr | both
invert = none | rts | dtr | both
```

Verified against a real UART, reading the lines back **through the same fd** — reading via a second `open()` cannot work, because opening a serial port re-asserts RTS/DTR, which had me briefly thinking the code was broken:

```
rts, normal              idle RTS=0        keyed RTS=1        unkeyed RTS=0
dtr, normal              idle DTR=0        keyed DTR=1        unkeyed DTR=0
both, invert rts (AIOC)  idle RTS=1 DTR=0  keyed RTS=0 DTR=1  unkeyed RTS=1 DTR=0
```

Each mode touches only its configured lines, and the idle state is now driven explicitly at open — that is the stuck-carrier case I mentioned, and it is the one worth having a test story for.

**New CM108 backend.** Keys a GPIO pin on CM108/CM119-class USB sound chips via a 5-byte HID report. The AIOC is in that table too (`1209:7388`), so it gives an AIOC user a second way in when the CDC port enumerates awkwardly. It talks to `/dev/hidraw` directly rather than linking hidapi — the report is five bytes and enumeration is a sysfs walk, so the dependency would only add a package to install on a Pi, and hamlib drives CM108 the same way. Linux-only, and it says so rather than failing obscurely. `cm108_ptt_report()` is exposed so the exact wire bytes are unit-tested without hardware; GPIO n is bit n-1 in bytes 2 **and** 3, and an off-by-one there keys the wrong pin invisibly.

**On naming**, I did rename `serial_rts` → `serial`, for the reason in my earlier comment — the old name bakes in a wire detail that is now one option among six. `serial_rts` is still accepted and means `line = rts`, so nobody's `mercury.ini` breaks. Happy to revert that specific bit if you would rather keep it.

**Also added the test we discussed**: `radio_model = 0` → `hermes_shm`. Plus the alias, the AIOC round-trip, defaults, invalid line/invert rejection, and the CM108 encoding. 13/13 in `test_cfg_utils`, full suite green, `go vet` clean, mercury builds clean.

**Not done:** line/invert/gpio controls in the three UIs. They select the method and `cm108` is offered, but the advanced serial options are ini-only. That is the obvious next piece and I did not want to sprawl further into your UI work without asking.

Two things from my earlier review still stand and are untouched here: `radio_io_init()` holding the radio mutex across `backend->open()` while reachable at runtime, and the legacy `radio_model = 0` leaving `hamlib_model = 0`.

ptsmonteiro:
Hi Rafael.
Used your commit and fixed a bunch of things on top of it.

In summary:

This branch expands and modernizes Mercury’s PTT support:
- Introduces a dedicated PTT configuration supporting none, hamlib, serial, cm108, and hermes_shm.
- Adds cross-platform serial PTT using RTS, DTR, or both, with independent line inversion.
- Adds CM108/CM119 GPIO PTT with automatic device discovery, optional hidapi, and Linux hidraw fallback.
- Exposes all PTT settings in both the Fyne and web interfaces, including runtime reconfiguration.
- Adds -P/-A CLI configuration and a -Q command to key PTT for one second as a hardware test.
- Preserves compatibility with legacy radio settings and the serial_rts method alias.
- Improves safety by explicitly parking PTT lines unkeyed during startup/shutdown and fixes CM108 unkey reporting and stale serial-device reuse.
- Updates the README, man page, example configuration, web documentation, and build configuration.
- Adds coverage for configuration parsing and round-tripping, legacy mappings, serial options, CM108 report encoding/device validation, and Fyne UI state handling.

rafael2k:
Thanks for picking those up — and your two additions are both improvements on what I handed over.

`cfg_ptt_config_set_method()` in particular fixes a real bug in mine: my `serial_rts` alias only mapped the *method*, so a stale `line = both` left in someone's ini would have silently combined with it. Forcing `line=rts, invert=none` makes the alias exact. And `-Q` is the right tool for this whole area — I ran it against a real UART with the AIOC config and it does exactly what an operator needs:

```
PTT ON for one second...
[INF] [radio-io] PTT method: serial RTS+DTR, inverted RTS (device /dev/ttyS0)
PTT OFF. Test complete.
```

**One gap left, and it was mine:** CM108 was effectively Linux-only, which does not match a project shipping Linux, Windows and macOS. `/dev/hidraw` exists only on Linux, and I had deliberately skipped `pkg-config` detection for Windows — correctly, since a cross-build's pkg-config finds the *host* hidapi and would link something that cannot run on the target — but I then left Windows with no path at all.

Fixed on **`ptt-hidapi-windows`** (commit `5c8f15e`), branched off your current head `a2b1b0d`:

| Platform | hidapi source |
|---|---|
| Linux | system via pkg-config, else `/dev/hidraw` (so a minimal Pi build still needs no extra package) |
| macOS | system via pkg-config — `brew install hidapi` |
| Windows | **vendored** in `radio_io/hidapi-w64` |

Vendored from **source** rather than a prebuilt library, unlike `hamlib-w64`: hidapi's Windows backend is self-contained over setupapi/cfgmgr32, so it compiles with the cross toolchain and there is no `.a` to keep in sync with a compiler version. 156 KB, hidapi 0.14.0, licence files included (GPLv3 is among its options, so compatible with ours).

Verified on the artifact rather than by reading the Makefile — the Windows binary links hidapi (`hid_enumerate` / `hid_open_path` / `hid_write` all present), carries the hidapi code path, and does **not** contain the "implemented for Linux only" stub string, so it genuinely took the real arm. Linux still builds, links `libhidapi-hidraw`, suite green.

Two traps I hit that are worth knowing, since both are silent:

- The vendored `hid.o` rule, placed near the detection block, became the Makefile's **default target** — `make windows` then "succeeded" after building only that one object. It has to sit below `all:`.
- `HIDAPI_CFLAGS` points at `radio_io/hidapi-w64`, which is right from the top level and wrong inside the `radio_io` sub-make. Exactly the same trap `hamlib-w64` has, now handled the same way with a `HIDAPI_CFLAGS_LOCAL`.

Cherry-pick whichever way suits you. With that in, #215 covers all three platforms for every method.

rafael2k:
`c2fb11d` is a good catch and the bug was mine. I checked it against hamlib's `src/cm108.c`, which is the reference for this report format, and you are right:

```c
(pttx == RIG_PTT_ON) ? (1 << ptt_bitnum) : 0,  // byte 2: GPIO output value
1 << ptt_bitnum,                               // byte 3: Data direction register (1=output)
```

Byte 3 has to keep the bit. Zeroing it on unkey turns the pin back into an input and leaves the output floating, so the radio can stay keyed — a stuck carrier, which is the one failure this backend must never have. Worse, my unit test asserted `r[3] == 0x00`, so it encoded the mistake as a passing test. A test that locks in the bug is worse than no test, and I had specifically claimed that one made the wire bytes safe without hardware.

For the record on where it came from: I ported the backend from modem73's `cm108_ptt.hh`, which has the same `buf[2] = 0; buf[3] = 0;` on unkey. I treated it as authoritative for the wire format when I should have read hamlib's encoding, having already cited hamlib as precedent for the transport. I've fixed it there too.

**Rebased the Windows work onto your corrected code**: `ptt-hidapi-windows` is now `1f2d481`, sitting directly on your `c812520`, so it carries the fixed encoding rather than my broken one. Force-pushed, so the old commit is gone.

Verified all three platforms from that base:

- **Linux** — builds, links `libhidapi-hidraw`, full suite green including your `test_cm108_report_encoding` and `test_cm108_linux_device_path_validation`
- **Windows** — cross-builds; the binary links hidapi (`hid_enumerate` / `hid_open_path` / `hid_write`), and does **not** contain the "implemented for Linux only" stub string, so it takes the real arm rather than the unsupported one
- **macOS** — system hidapi via pkg-config (`brew install hidapi`); unchanged by this commit and not buildable here

That is the last piece I know of. With it, every PTT method works on all three platforms we ship.

rafael2k:
**Not ready to merge — the macOS job is failing, and it is my bug.**

```
duplicate symbol '_cm108_ptt_set' in:
    radio_io/cm108_ptt.o
    radio_io/hamlib-macos/lib/libhamlib.a[arm64](cm108.o)
```

hamlib exports `cm108_ptt_set()` and `cm108_ptt_get()` from its own `src/cm108.c` — the very file I read to get the report encoding right — and I reused the names anyway.

What makes it worth a comment rather than a quiet fix: **it only breaks on macOS**, because that is the one platform where hamlib is linked *statically* (the 28 MB `libhamlib.a` we vendor). On Linux and Windows hamlib is a shared library, so the duplicate resolves silently at load time and which definition wins is a coin toss — our PTT calls could bind to hamlib's implementation, or its internals to ours, with no diagnostic on either side. So this was already latent on all three platforms and only the static link surfaced it.

Fixed on `ptt-hidapi-windows` (`4e63664`): renamed to `mercury_cm108_*`, with the reason in the header so it does not come back. Verified without a Mac, against the archive itself — `libhamlib.a` defines `_cm108_ptt_set`/`_cm108_ptt_get`, and `radio_io/cm108_ptt.o` now exports only `mercury_cm108_*`. Linux builds and the full suite is green; the Windows cross-build links clean.

That branch now carries two commits on top of your `c812520`: the Windows hidapi vendoring and this rename.

---

**On the rest of the review** — I went through the UI surface since it touches every client:

- All three UIs (Fyne, embedded `mercury.html`, PWA `docs/app/index.html`) expose the same five controls: method, device, line, invert, CM108 GPIO. I checked each by id rather than by eye; they are in sync. (My first pass reported the GPIO selector missing from the web UIs — that was a bad grep on my side, `cm108GPIOSelect` vs `cm108_gpio`. It is there.)
- `radio_list` is purely additive — `type`, `selected`, `device_path`, `serial_speed`, `list` all unchanged, with `ptt_method` / `ptt_line` / `ptt_invert` / `cm108_gpio` appended. Old clients keep parsing it.
- `set_radio_config` is still accepted, so **mercury-qt keeps working**.

**One hazard worth a decision, not a blocker.** `set_radio_config` unconditionally overwrites `config.method`, mapping the model number to none/hermes_shm/hamlib. An older client that sends it — mercury-qt, or a stale cached PWA — would silently switch an operator away from `serial` or `cm108` and persist it. Concretely: someone sets up an AIOC on `method = serial`, later opens an old mercury-qt, and their PTT stops working with nothing to explain why.

Options: ignore the legacy command when the current method is one it cannot express, or let it set only the hamlib fields and leave `method` alone unless the model actually changes. Either is a few lines. Worth deciding before this ships, since mercury-qt lives in a separate repo and will lag.

ptsmonteiro:
> One hazard worth a decision, not a blocker. set_radio_config unconditionally overwrites config.method, mapping the model number to none/hermes_shm/hamlib. An older client that sends it — mercury-qt, or a stale cached PWA — would silently switch an operator away from serial or cm108 and persist it. Concretely: someone sets up an AIOC on method = serial, later opens an old mercury-qt, and their PTT stops working with nothing to explain why.

Is this really a problem? We're shipping everything together so I don't think there will be "an older client".

Incorporated your commits from ptt-hidapi-windows. Tested the build on a real mac and it's working.

rafael2k:
Obrigado @ptsmonteiro !
@pedromessetti will take care of this PR, as this week I'll not be too available.

rafael2k:
Filed the legacy-command hazard as #219 so it is tracked independently of this PR — it will outlive the merge, and the client most likely to trigger it (mercury-qt) lives in another repo. Not a blocker here.

rafael2k:
@ptsmonteiro I've opened #220, which continues this work — it has this branch's 13 commits rebased onto current `mercuryv2` (post-1.9.13), plus one new commit adding CM108 hidapi support for macOS.

*(Correcting myself: I first wrote "your 14 commits", which was wrong twice over — this branch has 13, and 9 of them are yours, the other 4 mine.)*

Rather than decide unilaterally how to land it, which would you prefer?

1. **Close this in favour of #220.** Your commits are all in there with authorship intact — nine of the fourteen are yours — so nothing is lost, but this PR's review history stops here.
2. **Retarget #220 onto `feat/ptt-methods`**, so it becomes a PR against your branch and the additions land here as a follow-up. This PR stays the one that merges.
3. **I push the additions to `feat/ptt-methods` directly** and close #220, if you'd rather keep everything on your branch.

No preference from my side — whichever keeps the review sanest for you.

Two things from this branch worth your eye either way:

- **Your fix to the unkey report was right**, and my test was asserting the wrong value. Byte 3 is the direction register and has to keep carrying the GPIO bit when clearing the output, or the transmitter can stay keyed. Verified against hamlib's own implementation.
- **A collision that only shows on macOS**: hamlib exports `cm108_ptt_set`/`cm108_ptt_get`. Ours are now `mercury_cm108_*`. Invisible on Linux and Windows — it only bites where hamlib is linked statically.

And a question I can't answer myself: I tested on a macOS VM with no CM108 dongle attached, so enumeration and the error paths are verified but **keying a real radio is not** — neither on macOS nor Windows. Do you have a dongle you could point at a rig? That's the one gap before this is field-safe.

ptsmonteiro:
@rafael2k As this pull request is becoming too big I'd say it's better to merge it.  You can then rebase #220 and merge it later as something that was built on top of this.
I don't have any CM108 to test unfortunately.

PR #216, opened 2026-08-21
Support VarAC Beacon

pedromessetti:
VarAC sends its beacon on the broadcast port (8100) as a KISS `CMD_DATA` (0x02) "unformatted" frame. Mercury's `bcast_process_decoded_frame()` treated every `CMD_DATA` frame as a hermes-broadcast frame that already carries a Mercury header, so it did not wrap the beacon with the Mercury broadcast header. It just zero-padded 20→126 and put it on air.

**Fix:**

In `data_interfaces/tcp_interfaces.c`, distinguish the two `CMD_DATA` clients by content:
- **hermes-broadcast:** first byte already a broadcast header (`PACKET_TYPE_BROADCAST_CONTROL/_DATA`) → passed raw (unchanged).
- **VarAC beacon/ping:** no broadcast header → wrapped like `0x00`/`0x01` with the Mercury header + 2-byte length prefix, plus a new `BCAST_EXT_KISS_DATA` (0x04) extension bit so the far side mirrors the `0x02` framing and strips the padding.

`bcast_get_tx_payload()` now returns `CMD_DATA` for those length-prefixed frames, delivering the exact 20-byte payload instead of the padded frame.

This is symmetric, so beacons now work both directions, and existing `0x00`/`0x01` broadcast/Reticulum and hermes-broadcast paths are untouched.

### DIFF
--- a/data_interfaces/tcp_interfaces.c
+++ b/data_interfaces/tcp_interfaces.c
@@ -997,8 +997,9 @@ void *server_worker_thread_ctl(void *port)
  * header extension (BCAST_EXT_KISS_STD for 0x00, BCAST_EXT_KISS_DATA for 0x02,
  * neither for 0x01) so the receiving side can mirror it.
  *
- * hermes-broadcast (CMD_DATA with an existing Mercury broadcast header): the
- * frame is passed raw — zero-pad if short, discard if oversized.
+ * hermes-broadcast (CMD_DATA): a full modem frame (frame_len == frame_size)
+ * with a Mercury broadcast header already in place — passed raw.  Oversized
+ * CMD_DATA frames are discarded.
  *
  * Also latches bcast_reply_cmd so send_thread mirrors the client's framing.
  *
@@ -1014,18 +1015,33 @@ static bool bcast_process_decoded_frame(uint8_t *decoded_frame, int frame_len,
         memory_order_relaxed);
 
     /* CMD_DATA is used by two very different clients:
-     *   - hermes-broadcast: full modem frame with a Mercury broadcast header
+     *   - hermes-broadcast: a full modem frame with a Mercury broadcast header
      *     already in place (PACKET_TYPE_BROADCAST_CONTROL/_DATA) → pass raw.
-     *   - VarAC beacons/pings: unformatted KISS payload with no Mercury header
-     *     (its first byte is arbitrary and may even decode as an ARQ type) →
-     *     wrap it like an AX.25 frame so the receiver routes it to broadcast.
-     * Distinguish by the header byte already being a broadcast type. */
+     *   - VarAC beacons/pings: a short unformatted KISS payload with no Mercury
+     *     header, whose first byte is arbitrary (it can even land in the
+     *     broadcast-type range — e.g. a lowercase letter) → wrap it like an
+     *     AX.25 frame so the receiver routes it to broadcast.
+     * Distinguishing by the header bits alone misclassifies ~25% of arbitrary
+     * first bytes, so require the frame to also be exactly frame_size: a real
+     * hermes-broadcast frame always fills the modem frame, while a 20-byte
+     * beacon can never satisfy that. */
     bool is_raw_modem_frame = false;
     if (kiss_cmd == CMD_DATA)
     {
         uint8_t pt = frame_header_packet_type(decoded_frame[0]);
-        is_raw_modem_frame = (pt == PACKET_TYPE_BROADCAST_CONTROL ||
-                              pt == PACKET_TYPE_BROADCAST_DATA);
+        is_raw_modem_frame =
+            (pt == PACKET_TYPE_BROADCAST_CONTROL ||
+             pt == PACKET_TYPE_BROADCAST_DATA) &&
+            (size_t)frame_len == frame_size;
+    }
+
+    /* hermes-broadcast / unformatted CMD_DATA frames never fragment: anything
+     * larger than one modem frame is discarded (0x00/0x01 truncate instead). */
+    if (kiss_cmd == CMD_DATA && (size_t)frame_len > frame_size)
+    {
+        HLOGW("tcp-bcast", "Discarding broadcast frame: size %d exceeds modem frame size %zu",
+              frame_len, frame_size);
+        return false;
     }
 
     bool needs_wrap = (kiss_cmd == CMD_AX25 || kiss_cmd == CMD_AX25CALLSIGN) ||
--- a/data_interfaces/tcp_interfaces.c
+++ b/data_interfaces/tcp_interfaces.c
@@ -1015,16 +1015,33 @@ static bool bcast_process_decoded_frame(uint8_t *decoded_frame, int frame_len,
         memory_order_relaxed);
 
     /* CMD_DATA is used by two very different clients:
-     *   - hermes-broadcast: a full modem frame with a Mercury broadcast header
+     *   - hermes-broadcast: a modem frame with a Mercury broadcast header
      *     already in place (PACKET_TYPE_BROADCAST_CONTROL/_DATA) → pass raw.
-     *   - VarAC beacons/pings: a short unformatted KISS payload with no Mercury
-     *     header, whose first byte is arbitrary (it can even land in the
-     *     broadcast-type range — e.g. a lowercase letter) → wrap it like an
-     *     AX.25 frame so the receiver routes it to broadcast.
-     * Distinguishing by the header bits alone misclassifies ~25% of arbitrary
-     * first bytes, so require the frame to also be exactly frame_size: a real
-     * hermes-broadcast frame always fills the modem frame, while a 20-byte
-     * beacon can never satisfy that. */
+     *   - VarAC beacons/pings: an unformatted KISS payload with no Mercury
+     *     header, whose first byte is arbitrary → wrap it like an AX.25 frame
+     *     so the receiver routes it to broadcast.
+     *
+     * The type bits alone are not enough: they are the top 3 of the first
+     * byte, so 64 of 256 arbitrary first bytes look like a broadcast type --
+     * 25%, including the whole lowercase alphabet.  A beacon starting with a
+     * lowercase letter would be passed raw and lost.
+     *
+     * Frame LENGTH cannot be the second test either, tempting as it is.
+     * hermes-broadcast does not always fill the modem frame: transmitter.c
+     * sends payload frames as packet_size + RQ_HEADER_SIZE (== frame_size) but
+     * CONFIG frames as packet_size alone, i.e. RQ_HEADER_SIZE short, and
+     * relies on the zero-padding below to bring them up.  Requiring
+     * frame_len == frame_size drops every config packet, and since its
+     * receiver discards anything that is not exactly frame_size, the far end
+     * never gets the RaptorQ configuration and can never start decoding.
+     *
+     * So validate the WHOLE header byte instead.  hermes-broadcast writes both
+     * of its frame kinds with extension 0 (hermes_write_frame_header(...,
+     * PACKET_RQ_CONFIG/PACKET_RQ_PAYLOAD, 0)), while a beacon's extension bits
+     * are as arbitrary as the rest of its first byte.  Type plus a zero
+     * extension takes the false-positive space from 64/256 to 2/256 (0.8%),
+     * and to ZERO across a-z: the only printable ASCII that still collides is
+     * a leading backtick. */
     bool is_raw_modem_frame = false;
     if (kiss_cmd == CMD_DATA)
     {
@@ -1032,7 +1049,7 @@ static bool bcast_process_decoded_frame(uint8_t *decoded_frame, int frame_len,
         is_raw_modem_frame =
             (pt == PACKET_TYPE_BROADCAST_CONTROL ||
              pt == PACKET_TYPE_BROADCAST_DATA) &&
-            (size_t)frame_len == frame_size;
+            frame_header_extension(decoded_frame[0]) == 0;
     }
 
     /* hermes-broadcast / unformatted CMD_DATA frames never fragment: anything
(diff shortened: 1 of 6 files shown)


rafael2k:
The framing work is right and the symmetry argument is sound — a beacon that goes out wrapped comes back unwrapped, and the `0x00`/`0x01` and hermes-broadcast paths are genuinely untouched. CI is green and the two new tests are the right shape.

**One thing needs addressing before merge: the discrimination is content-based, and it misfires on ordinary beacon text.**

```c
uint8_t pt = frame_header_packet_type(decoded_frame[0]);
is_raw_modem_frame = (pt == PACKET_TYPE_BROADCAST_CONTROL ||
                      pt == PACKET_TYPE_BROADCAST_DATA);
```

`frame_header_packet_type()` is the **top 3 bits**, so this claims "already has a Mercury header" for any payload whose *first byte* happens to land in:

| packet type | first-byte range | what that is in ASCII |
|---|---|---|
| `BROADCAST_CONTROL` (0x03) | `0x60–0x7F` | **`` ` `` and `a`–`z` and `{ \| } ~`** |
| `BROADCAST_DATA` (0x04) | `0x80–0x9F` | high-bit bytes |

That is 64 of 256 first bytes — **25% of arbitrary payloads**, and it includes the entire lowercase alphabet. A beacon body beginning with a lowercase letter is treated as a raw hermes-broadcast frame, passed through unwrapped, and lost.

I added a throwaway probe to your branch to check rather than argue from the constants:

```
  first byte 0x20 (type 1) -> wrapped (correct)
  first byte 0x43 (type 2) -> wrapped (correct)     'C'
  first byte 0x61 (type 3) -> PASSED RAW (beacon lost)   'a'
  first byte 0x7A (type 3) -> PASSED RAW (beacon lost)   'z'
  first byte 0x85 (type 4) -> PASSED RAW (beacon lost)
```

Your `test_bcast_rx_cmd_data_varac_beacon_wrapped` uses `0x20` as the first byte, which is type 1 and therefore always on the safe side of the branch — so the test cannot catch this. Its comment ("first byte 0x2E decodes as an ARQ type") also does not match the code, which writes `0x20`.

Whether it bites today depends entirely on what VarAC actually puts first in a beacon. If that is always a fixed marker or an uppercase callsign, this never fires in practice — but it is fragile by construction and fails silently, and "the first byte happened to be lowercase" is a miserable bug to chase off-air.

**Suggestions**, in order of preference:

1. **Do not infer from content.** The two CMD_DATA clients are distinguishable at connect time, not per frame — hermes-broadcast is a Mercury-aware client, VarAC is not. If a client can be latched as one or the other on its first frame (or by a handshake/flag), the guess disappears.
2. **Tighten the test.** A real hermes-broadcast frame is a *full* modem frame — `frame_len == frame_size`. A VarAC beacon is short and length-prefixed. Requiring `frame_len == (int)frame_size` **and** a broadcast packet type would cut the false-positive rate enormously, since a 20-byte beacon can never satisfy the first condition.
3. If the heuristic must stay as-is, at least extend the test to cover a first byte in `0x60–0x7F` so the limitation is visible and deliberate rather than accidental.

(2) looks like a one-line change and closes the realistic case.

Everything else — the `BCAST_EXT_KISS_DATA` bit choice, the mutual exclusivity with `_KISS_STD`, the receive-side mirroring, the docs — reads correctly to me.

rafael2k:
Re-reviewed `392a9e4`. The fix is right and I verified it rather than reading it — re-ran the same probe I used to find the problem, now across the whole collision range:

```
  short beacon, first 0x20 (type 1) -> wrapped (correct)
  short beacon, first 0x43 (type 2) -> wrapped (correct)
  short beacon, first 0x61 (type 3) -> wrapped (correct)   'a'
  short beacon, first 0x7A (type 3) -> wrapped (correct)   'z'
  short beacon, first 0x85 (type 4) -> wrapped (correct)
  short beacon, first 0x9F (type 4) -> wrapped (correct)
  full-size hermes-broadcast        -> passed raw (correct)
  beacons lost: 0 of 6
```

0 of 6 lost, where it was 3 of 5 before, and a genuine full-size hermes-broadcast frame still passes raw. The two new tests cover exactly the cases that were missing, including the lowercase one. Full suite green here. Good fix, and the docs and comments now say *why* the length check exists rather than just what it does.

**One behaviour change to confirm is intended**, because it is the mirror of the bug and I do not think it is covered.

A **short** hermes-broadcast frame — valid Mercury broadcast header, `frame_len < frame_size` — is now wrapped rather than passed raw. Probed on this branch:

```
  short hermes-broadcast (3 of 10 bytes):
    out[0]=0x85 out[1]=0x00 out[2]=0x03
    -> WRAPPED (behaviour changed)
```

That is the wrap header plus the 2-byte length prefix, with the client's original Mercury header now sitting inside as payload.

It matters because the pre-PR contract explicitly allowed short frames here — the old comment read *"hermes-broadcast (CMD_DATA): the frame already carries the Mercury header; **zero-pad if short**, discard if oversized"*, and `docs/TNC.md` said the same. So short frames were anticipated, not impossible.

The end result may well be harmless or better: the receiver strips the wrapper and delivers the exact original bytes instead of a zero-padded frame, and `BCAST_EXT_KISS_DATA` still makes it reply with `0x02`. But it is a change in what a hermes-broadcast peer receives — exact length instead of a padded full frame — and it is currently neither documented nor tested.

Worth one of:

- confirming hermes-broadcast always sends full frames, in which case delete the "zero-pad if short" language that is now stale and add a test pinning that a short one is wrapped deliberately; or
- keeping short raw frames raw, by treating a broadcast header as authoritative when the frame is short *and* the length check only as a tiebreak — though that reopens the lowercase case, so I would not.

My preference is the first: the new behaviour looks fine, it just should not be accidental. Everything else here is good to go from my side.

rafael2k:
**Please hold this — I traced the short-frame case into hermes-broadcast and it does break it.** I raised it as "confirm this is intended"; it is not intended.

hermes-broadcast sends its two frame kinds at *different lengths*, and only one of them fills the modem frame (`transmitter.c`, `packet_size = frame_size - RQ_HEADER_SIZE`):

```c
/* payload */ tcp_interface_send_kiss(&tcp_iface, data, packet_size + RQ_HEADER_SIZE);
/* config  */ tcp_interface_send_kiss(&tcp_iface, full_packet, packet_size);
```

For hermes mode 1 (`hermes_frame_size[1] = 126`, `RQ_HEADER_SIZE = 4`):

| frame | bytes sent | `== frame_size`? | under this PR |
|---|---|---|---|
| RaptorQ payload | 122 + 4 = **126** | yes | passed raw — correct |
| **config packet** | **122** | **no** | **wrapped as if it were a beacon** |

And the config packet does carry a Mercury broadcast header — `PACKET_RQ_CONFIG` is `0x03`, which is exactly `PACKET_TYPE_BROADCAST_CONTROL`. So it satisfies the type test and fails only the new length test.

The other end then rejects it, because hermes-broadcast's receiver is strict about length (`receiver.c`):

```c
if ((uint32_t)frame_len == frame_size) { ...accept... }
fprintf(stderr, "Discarding unexpected TCP frame length %d (expected %u)\n",
        frame_len, frame_size);
return 0;
```

So the wrapped config packet is unwrapped by the receiving Mercury, delivered as **122** bytes, and discarded as the wrong length. Before this change it was zero-padded to 126 and accepted — which is precisely what the old *"hermes-broadcast (CMD_DATA): zero-pad if short"* contract was doing. That padding was load-bearing, not incidental.

The symptom in the field would be: RaptorQ payload frames flow, the config packet never arrives, and the receiver can never start decoding — with one line in its stderr as the only clue.

**Suggested fix**, keeping the beacon fix intact: the length test should be "does not exceed the frame" rather than "exactly fills it", since the real discriminator is that hermes-broadcast frames carry a genuine Mercury broadcast header *and* are padded up by Mercury anyway:

```c
is_raw_modem_frame =
    (pt == PACKET_TYPE_BROADCAST_CONTROL || pt == PACKET_TYPE_BROADCAST_DATA) &&
    (size_t)frame_len >= HEADER_SIZE_OF_A_REAL_BROADCAST_FRAME;
```

That alone reopens the lowercase case, so it needs a second discriminator. Two that would work:

1. **Latch per client.** hermes-broadcast is Mercury-aware and VarAC is not; whichever a client is, it does not change mid-connection. Deciding once on the first frame removes the per-frame guess entirely — this was my first suggestion and it is looking like the right one.
2. **Validate the whole header, not just the type bits.** A real hermes-broadcast frame's extension bits are meaningful (`PACKET_RQ_CONFIG`/`PACKET_RQ_PAYLOAD` with `extension = 0`), whereas a beacon's first byte is arbitrary. Checking type *and* that the extension is zero cuts the false-positive space by 32x, and a short config packet still passes.

Either keeps VarAC beacons working without breaking hermes-broadcast's config packet.

Sorry to move the goalposts — the frame-length idea was mine, and I did not check hermes-broadcast's sender before suggesting it. The beacon half of this PR is right; it is only the discriminator that needs another pass.

rafael2k:
Pushed the fix to this branch as `e2c500a`, since the frame-length idea was mine and it was the thing that broke hermes-broadcast.

**The discriminator is now the whole first byte, not the length.** hermes-broadcast writes both of its frame kinds with extension 0 (`hermes_write_frame_header(..., PACKET_RQ_CONFIG / PACKET_RQ_PAYLOAD, 0)`), while a beacon's extension bits are as arbitrary as the rest of its first byte:

| test | false positives |
|---|---|
| type bits only (the original bug) | 64/256 — **25%**, all of `a`–`z` |
| type + zero extension (now) | 2/256 — **0.8%**, none of `a`–`z` |

So beacons are still fixed — the lowercase case that started this is covered by a new test — and short config frames go back to being passed raw and zero-padded.

**One test of yours I rewrote rather than deleted.** `test_bcast_rx_cmd_data_short_wrapped` pinned `0x60`, which is the single genuinely ambiguous byte: `BROADCAST_CONTROL` with a zero extension is byte-for-byte a hermes-broadcast config header *and* a printable backtick. Nothing in the frame separates them. I resolved it toward hermes-broadcast and renamed the test to `test_bcast_rx_cmd_data_ambiguous_0x60_treated_as_hermes`, with the reasoning in the comment: hermes-broadcast is in daily use and its stream dies silently if we guess wrong, whereas a beacon whose payload starts with a backtick is hypothetical. If you disagree, that is the line to argue about — everything else follows from it.

Also added two regression tests: a short config frame stays raw and padded, and a short beacon starting with `a` (type 3, extension 1) is still wrapped.

Full suite green, 64/64 in `test_tcp_interfaces`, clean build.

---

**Separately, hermes-broadcast has its own bug here**, which is what made this fragile in the first place. Its TCP config send says one thing and does another:

```c
// For TCP, send full frame with stuffing
tcp_interface_send_kiss(&tcp_iface, full_packet, packet_size);   // frame_size - RQ_HEADER_SIZE
```

while the payload path right above sends `packet_size + RQ_HEADER_SIZE`. So config frames were 4 bytes short of every other frame and only arrived because Mercury padded them — and its own receiver then requires exactly `frame_size`. Fixed in `rafael2k/hermes-broadcast` (`009f4c1`, committed locally, builds clean): the config path now emits a full frame instead of relying on the transport to repair its framing.

Mercury keeps the padding regardless, so that fix is not required for this PR to be correct — the two are independent.

rafael2k:
Pushed `a5dc1a6`: raises the broadcast KISS payload cap from 756 to 1213.

While removing the dead Mercury-v1 shared-memory path from hermes-broadcast I ran every mode in the table end-to-end, and two of them had never worked.

**Modes 9 and 10 were unreachable.** `MAX_PAYLOAD` was 756, below DATAC17 (1180) and QAM16C2 (1213), so `send_thread`/`recv_thread` rejected them with *"Invalid broadcast frame size"*. On the receive side it was worse than a rejection: `kiss_read()` silently stops storing at `MAX_PAYLOAD`, so an oversized frame is **truncated**, not dropped — it arrives as 756 bytes and then fails the length check for the wrong reason. The two highest-rate modes were simply unavailable on the broadcast plane even though the modem offers them.

This is a buffer bound, not a wire format — nothing about the framing changes and frames at the old sizes are untouched. Full test suite passes.

The matching change is in hermes-broadcast (`ae305b8`); both sides are needed. With them, a 200 kB file round-trips byte-identical through a KISS/TCP relay in modes 0–4 and 6–10.

Two things worth flagging separately:

- **Mode 5 (DATAC14) cannot carry broadcast at all.** `frame_size` is 3, so `packet_size = frame_size - RQ_HEADER_SIZE` (3 − 4) underflows `size_t`, and the config packet copies 9 bytes into a 3-byte buffer. Fixed on the hermes-broadcast side by rejecting any mode whose frame can't hold the RaptorQ tag plus the config packet. Mercury itself is fine here — it never computes that subtraction — but it will happily accept a `broadcast_frame_size_cfg` of 3 from a client that does.
- **I introduced a regression in hermes-broadcast `009f4c1` and have reverted it** (`210b1cc`). I'd read `write_configuration_packet()`'s parameter — named `packet_size` — as the payload budget, but the caller passes `frame_size`, so my `+ RQ_HEADER_SIZE` made every config frame 4 bytes too long and the receiver discarded all of them. That is exactly the failure the commit claimed to fix; the pre-existing code was correct. Measured on the wire, mode 1: before 126 ✅, my commit 130 ❌, now 126 ✅. It only ever lived on hermes-broadcast `main`, never here.

If you'd rather keep this PR to the CMD_DATA/VARA scope, say so and I'll move `a5dc1a6` to its own branch.

rafael2k:
Disregard the last line of my previous comment — decision is to **keep `a5dc1a6` on this branch**, so it all merges together.

LGTM from my side. Four commits, full test suite green, ready to merge.

@ptsmonteiro one thing to keep in mind when you do: `a5dc1a6` only takes effect on air together with hermes-broadcast `ae305b8` — the KISS payload cap has to be raised on *both* ends, or DATAC17 and QAM16C2 stay unreachable on the broadcast plane.

PR #220, opened 2026-08-24
PTT method configuration: serial, CM108/AIOC, and three-platform hidapi

rafael2k:
Rebased onto `mercuryv2` now that #215 has merged. The 13 shared commits dropped automatically (identical patch-ids), so this PR is down to two things:

### 1. macOS CM108 support — `80d3aba`

#215 vendored hidapi for Windows and used pkg-config elsewhere. On macOS there is no system hidapi to find, so `HAVE_HIDAPI` was never defined and the CM108 path silently compiled out: the PTT method would appear in the UI and then do nothing. This vendors the hidapi 0.14.0 IOKit backend the same way `hidapi-w64` is vendored, so all three platforms ship the same feature.

`radio_io/include/hidapi.h` is byte-identical to the w64 copy. The Makefile prefers the vendored tree on Darwin and keeps pkg-config as the fallback.

### 2. Web UI field visibility parity with Fyne — `bda642c`

The Fyne PTT dialog hides the rows a method does not use. Both web UIs only set `.disabled`, so all six controls stayed on screen greyed out — and a `<label>` has no disabled state, so "Hamlib Baud Rate", "PTT Line", "Invert Line" and "CM108 GPIO" kept rendering at full contrast beside a dead control. To an operator that reads as "this applies to me but is broken", which is the opposite of what the Fyne user sees.

Both `gui_interface/websocket/web/mercury.html` and `docs/app/index.html` now drive visibility off the same matrix as the Fyne switch:

| method | fields shown |
|---|---|
| `hamlib` | device path, model, baud |
| `serial` | device path, PTT line, invert |
| `cm108` | device path, GPIO |
| `none`, `hermes_shm` | none |

Verified in Chrome against the served page: the visible-row set matches the Fyne matrix for all five methods, and hidden controls report `disabled` (so they are not keyboard-reachable or announced).

Two parity gaps turned up while comparing the three UIs side by side, both fixed here:

- **`docs/app/index.html` never got the stale-path guard.** Switching to CM108 from serial or hamlib left `/dev/ttyUSB0` in the device field, handing a serial path to the HID opener. mercury.html and Fyne both clear it.
- **`mercury.html`'s `setSelectValue()` returned nothing**, but the baud-rate handler tests its result: `if (!setSelectValue(sel, baud))`. `undefined` is always falsy, so a reported speed that *did* match an option was immediately overwritten with "0" (Auto) on every status message. `docs/app/index.html` already had the two-line fix; carried across.

### Testing

Native build clean; `make -C tests test` 278 assertions, 0 failures.

Still untested by anyone: keying a real radio over CM108. Neither ptsmonteiro nor I have the hardware, so that remains open on all three platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,7 @@ HIDAPI_CFLAGS =
 HIDAPI_LDFLAGS =
 HIDAPI_OBJS =
 HIDAPI_W64_DIR = radio_io/hidapi-w64
+HIDAPI_MACOS_DIR = radio_io/hidapi-macos
 ifeq ($(OS),Windows_NT)
   # Vendored, like hamlib-w64: pkg-config on a cross-build would find the HOST
   # hidapi and link something that cannot run on the target.  hidapi's Windows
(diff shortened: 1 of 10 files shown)


rafael2k:
Asked @ptsmonteiro in #215 how he'd like this landed — closing #215 in favour of this, retargeting this onto `feat/ptt-methods`, or pushing the additions to his branch instead. Holding off on merging either until he says.

PR #221, opened 2026-08-25
fix(arq): add bounded random jitter to retry deadlines

### LINKED ISSUE #217: Retry needs randomness
## What did we do?

We were two radio amateurs sitting next to each other, both having `mercury` up and running on their PC. There were no radios involved. We operated from the speakers of one PC to the microphone to the other and reverse, in order to play with mercury at ease.

We had established a connection. We were able to exchange messages by sending stuff into port 8301 on one side and extracting the data from port 8301 on the other side.

We then had the idea to count down and then submit data to port 8301, to be transported to the other side, simultaneously on both sides. 4 - 3 - 2 - 1 - go.

## Expectation

After some initial confusion, the two instances negotiate who is to send. Eventually, the messages from both ends reach their respective destinations.

## Actually seen

Both instances send exactly simultaneously. They continue to do so. As they can't receive while sending, no instance ever gets to hear anything the other sends. In they end, they decide on "disconnect". No data ever reaches the other end.

## Our humble opinion / suggestion

Network stuff needs a measure of randomness. If retries are done according to a fixed schedule, stuff like this can happen. Retries should not happen after a *fixed* time, but after some *fixed* time *plus some random additional delay* that is large enough so that with high probability, one of the instances starts hearing the other.

pedromessetti:
## Summary

Fixes #217 — two stations on a fixed retry schedule can phase-lock: both transmit simultaneously, neither can hear the other, and the session stalls until disconnect.

The fix adds a **deterministic per-node retry stagger**: the two peers compare the same pair of exchanged callsigns, so exactly one defers by a fixed offset on every retry. No random draw, no latency tail, and deadlines become exactly testable.

## How it works

```
rank     = strcmp(local_call, remote_call) > 0 ? 1 : 0
deadline = now + base + rank * retry_stagger_ms   /* default 2000 ms */
```

`strcmp()` is antisymmetric, so both ends compute the *same* comparison from the *same* pair and always disagree about who defers. The stagger accumulates across retransmits (each deadline is re-anchored at PTT-OFF), so the two stations drift apart deterministically and one soon transmits while the other is listening.

- Applied to all retry timers: CALL, ACCEPT, TURN_REQ, KEEPALIVE, DISCONNECT, MODE_REQ, and **the DATA retransmit (`ack_timeout_s`)** — the load-bearing timer for #217's exact scenario, where two connected stations both land in `WAIT_ACK`. `ack_timeout_s` is a lower bound, so a positive offset only ever delays the retransmit.
- `arq_protocol_retry_rank()` returns `-1` for a degenerate pair (identical or missing callsigns — a misconfiguration) and falls back to a random draw over `[0, stagger_ms)`.

## Config

- `arq:retry_stagger_ms` INI key (default 2000, range 0..5000, **0 disables**), plumbed like the neighbouring ARQ tunables.

## Tests

- `test_arq_protocol`: rank antisymmetry + determinism, degenerate-pair fallback, exact stagger deadlines, disabled-stagger exactness, bounded fallback.
- `test_arq_fsm`: PTT-OFF re-anchor and `WAIT_ACK` deadlines assert **exact** equality again.
- `test_cfg_utils`: default, off-switch (`0`) round-trip, out-of-range rejection.
- Full suite passes (23 binaries, 0 failures).

## Trade-offs

Deterministic vs random backoff: point-to-point ARQ has identity, so it doesn't need luck to break symmetry. The stagger is reproducible (a bad run replays exactly), has no tail latency (one side pays a fixed, known offset), and restores exact-equality test assertions. Random backoff remains only as a fallback for the same-callsign misconfiguration.

### DIFF
--- a/common/cfg_utils.c
+++ b/common/cfg_utils.c
@@ -65,6 +65,7 @@ void cfg_set_defaults(mercury_config *cfg)
     cfg->keepalive_miss_limit        = ARQ_KEEPALIVE_MISS_LIMIT_DEFAULT;
     cfg->peer_payload_hold_s         = ARQ_PEER_PAYLOAD_HOLD_S_DEFAULT;
     cfg->startup_max_s               = ARQ_STARTUP_MAX_S_DEFAULT;
+    cfg->retry_jitter_ms             = ARQ_RETRY_JITTER_MS_DEFAULT;
     cfg->busy_detect                 = false;
     cfg->busy_threshold_db           = 10;
     cfg->busy_hysteresis_db          = 3;
@@ -223,6 +224,9 @@ bool cfg_read(mercury_config *cfg, const char *ini_path)
     i = iniparser_getint(ini, CFG_KEY_ARQ_STARTUP_MAX_S, cfg->startup_max_s);
     if (i >= 2 && i <= 60)   cfg->startup_max_s = i;
 
+    i = iniparser_getint(ini, CFG_KEY_ARQ_RETRY_JITTER_MS, cfg->retry_jitter_ms);
+    if (i >= 0 && i <= ARQ_RETRY_JITTER_MS_MAX) cfg->retry_jitter_ms = i;
+
     cfg->busy_detect = (bool) iniparser_getboolean(ini, CFG_KEY_BUSY_DETECT,
                                                    cfg->busy_detect ? 1 : 0);
 
@@ -342,6 +346,7 @@ bool cfg_write(const mercury_config *cfg, const char *ini_path)
     fprintf(f, "keepalive_miss_limit = %d\n", cfg->keepalive_miss_limit);
     fprintf(f, "peer_payload_hold_s = %d\n", cfg->peer_payload_hold_s);
     fprintf(f, "startup_max_s = %d\n", cfg->startup_max_s);
+    fprintf(f, "retry_jitter_ms = %d\n", cfg->retry_jitter_ms);
 
     fprintf(f, "\n[channel]\n");
     fprintf(f, "busy_detect = %s\n", cfg->busy_detect ? "true" : "false");
--- a/common/cfg_utils.c
+++ b/common/cfg_utils.c
@@ -65,7 +65,7 @@ void cfg_set_defaults(mercury_config *cfg)
     cfg->keepalive_miss_limit        = ARQ_KEEPALIVE_MISS_LIMIT_DEFAULT;
     cfg->peer_payload_hold_s         = ARQ_PEER_PAYLOAD_HOLD_S_DEFAULT;
     cfg->startup_max_s               = ARQ_STARTUP_MAX_S_DEFAULT;
-    cfg->retry_jitter_ms             = ARQ_RETRY_JITTER_MS_DEFAULT;
+    cfg->retry_stagger_ms            = ARQ_RETRY_STAGGER_MS_DEFAULT;
     cfg->busy_detect                 = false;
     cfg->busy_threshold_db           = 10;
     cfg->busy_hysteresis_db          = 3;
@@ -224,8 +224,8 @@ bool cfg_read(mercury_config *cfg, const char *ini_path)
     i = iniparser_getint(ini, CFG_KEY_ARQ_STARTUP_MAX_S, cfg->startup_max_s);
     if (i >= 2 && i <= 60)   cfg->startup_max_s = i;
 
-    i = iniparser_getint(ini, CFG_KEY_ARQ_RETRY_JITTER_MS, cfg->retry_jitter_ms);
-    if (i >= 0 && i <= ARQ_RETRY_JITTER_MS_MAX) cfg->retry_jitter_ms = i;
+    i = iniparser_getint(ini, CFG_KEY_ARQ_RETRY_STAGGER_MS, cfg->retry_stagger_ms);
+    if (i >= 0 && i <= ARQ_RETRY_STAGGER_MS_MAX) cfg->retry_stagger_ms = i;
 
     cfg->busy_detect = (bool) iniparser_getboolean(ini, CFG_KEY_BUSY_DETECT,
                                                    cfg->busy_detect ? 1 : 0);
@@ -346,7 +346,7 @@ bool cfg_write(const mercury_config *cfg, const char *ini_path)
     fprintf(f, "keepalive_miss_limit = %d\n", cfg->keepalive_miss_limit);
     fprintf(f, "peer_payload_hold_s = %d\n", cfg->peer_payload_hold_s);
     fprintf(f, "startup_max_s = %d\n", cfg->startup_max_s);
-    fprintf(f, "retry_jitter_ms = %d\n", cfg->retry_jitter_ms);
+    fprintf(f, "retry_stagger_ms = %d\n", cfg->retry_stagger_ms);
 
     fprintf(f, "\n[channel]\n");
     fprintf(f, "busy_detect = %s\n", cfg->busy_detect ? "true" : "false");
--- a/common/cfg_utils.h
+++ b/common/cfg_utils.h
@@ -57,6 +57,7 @@
 #define CFG_KEY_ARQ_KEEPALIVE_MISS_LIMIT      "arq:keepalive_miss_limit"
 #define CFG_KEY_ARQ_PEER_PAYLOAD_HOLD_S       "arq:peer_payload_hold_s"
 #define CFG_KEY_ARQ_STARTUP_MAX_S             "arq:startup_max_s"
+#define CFG_KEY_ARQ_RETRY_JITTER_MS           "arq:retry_jitter_ms"
 #define CFG_KEY_BUSY_DETECT                   "channel:busy_detect"
 #define CFG_KEY_BUSY_THRESHOLD_DB             "channel:busy_threshold_db"
 #define CFG_KEY_BUSY_HYSTERESIS_DB            "channel:busy_hysteresis_db"
@@ -107,6 +108,9 @@ typedef struct {
                                          * activity. Default 15, clamped 1..120.*/
     int      startup_max_s;             /* ARQ: control-mode-only startup
                                          * window. Default 10, clamped 2..60.   */
+    int      retry_jitter_ms;           /* ARQ: bounded random delay added to
+                                         * retry deadlines. 0 disables. Default
+                                         * 2000, clamped 0..5000.              */
     float    tx_gain_db;            /* Linear-equivalent gain on the modulator
                                     * TX samples, in dB. 0.0 = no change.
                                     * Range -20.0 .. +20.0 (clamped). */
--- a/common/cfg_utils.h
+++ b/common/cfg_utils.h
@@ -57,7 +57,7 @@
 #define CFG_KEY_ARQ_KEEPALIVE_MISS_LIMIT      "arq:keepalive_miss_limit"
 #define CFG_KEY_ARQ_PEER_PAYLOAD_HOLD_S       "arq:peer_payload_hold_s"
 #define CFG_KEY_ARQ_STARTUP_MAX_S             "arq:startup_max_s"
-#define CFG_KEY_ARQ_RETRY_JITTER_MS           "arq:retry_jitter_ms"
+#define CFG_KEY_ARQ_RETRY_STAGGER_MS          "arq:retry_stagger_ms"
 #define CFG_KEY_BUSY_DETECT                   "channel:busy_detect"
 #define CFG_KEY_BUSY_THRESHOLD_DB             "channel:busy_threshold_db"
 #define CFG_KEY_BUSY_HYSTERESIS_DB            "channel:busy_hysteresis_db"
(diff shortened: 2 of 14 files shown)


rafael2k:
Reviewed by building the branch and measuring it against trunk, not just reading the diff. The mechanism is right and the implementation is careful — one substantive question about coverage, then some smaller things.

## Verified independently

- Builds clean; **270 unit assertions, 0 failures**, matching your claim.
- **The lock discipline is correct.** `retry_deadline_from_s()` calls `arq_conn_get_calls()`, which takes `g_conn_lock` from inside the FSM. That is safe here: `arq.c` documents `g_conn_lock` as a deliberate *leaf* lock with "no ordering relationship to maintain", and no path holds it across an FSM dispatch. Worth saying out loud because it is the kind of thing that deadlocks six months later.
- **Deterministic under the virtual clock.** The jitter derives from `time_now_ms()`, so sim runs stay reproducible — which is why the seeded tests still pass.

The per-node salt is the crux and it is done right: mixing the **local** callsign means two peers sharing a `session_id` and a synchronised clock still derive different jitter. That is exactly what #217 needs.

## The one substantive question: does this cover the timer that actually locks?

#217's scenario is two connected stations both submitting data on a countdown — simultaneous **DATA**, not simultaneous CALL. Each transmits, neither can hear, each enters `WAIT_ACK`, and each retransmits when `ack_timeout_s` expires.

That deadline is the one you deliberately excluded:

```c
dflow_enter(sess, ARQ_DFLOW_WAIT_ACK,
            deadline_from_s(tm ? tm->ack_timeout_s : 9.0f),   // arq_fsm.c:1674
```

Both stations enter `WAIT_ACK` at the same instant (they finished transmitting at the same instant), so their retransmits stay aligned on a fixed schedule — the exact lock the issue describes. Jittering CALL/ACCEPT/TURN_REQ does not help here, because while both are transmitting neither *receives* anything, TURN_REQ included; the only thing that can break the symmetry is the timer they are both sitting on.

If that reading is right, the session stays locked until the data retries exhaust and it disconnects — which is precisely the reported outcome.

**Suggestion, and I think it is safe:** jitter that deadline too. The "carefully tuned margin" on `ack_timeout_s` is a *lower* bound — it must be long enough for the peer's ACK to arrive. A bounded *positive* jitter only ever waits longer, so it cannot violate the margin; it only delays the retransmit. That makes it the one place where jitter is both safe and load-bearing for this issue.

Happy to be wrong about the failure path — you have talked to the reporters and I have not. But it is worth confirming against #217's actual scenario before this closes the issue.

## Latency cost: measured, and smaller than it first looked

Symmetric ITU-R `good` fading, Go harness, only the binary varies:

| cell | trunk | this PR |
|---|---|---|
| 1 KB @ 5.2 dB | 100 / 101 / 100 s | 130 / 102 / 100 s |
| 102 B @ −4.8 dB | 246 s | 247 s |

My first run showed 130 s against trunk's 100 s and I was about to report a 30 % regression — the repeats corrected that. The median cost is **zero**; what jitter adds is a *tail*, where one draw in three happened to cost a retry cycle. That is the expected shape for randomised backoff and looks like a fair trade, but it is worth a line in the PR body so nobody is surprised by a slow run later.

## Smaller points

1. **The two divergence assertions are fragile.** `TEST_ASSERT_NOT_EQUAL_UINT64(first, second)` and the salt test both require two *particular* hash draws to differ. With jitter uniform over [0, 2000) two draws collide about 1 time in 2000 — deterministic for today's inputs, so not flaky now, but it becomes a baffling failure the moment anyone changes the hash constants, the default jitter, or the timestamps in the test. Consider asserting the bounds plus divergence *over N samples* (e.g. ≥ K distinct values across 100 instants) rather than requiring one specific pair to differ.

2. **No runtime off-switch.** `ARQ_RETRY_JITTER_MS` is compile-time with no `.ini` plumbing, while its neighbours (`retry_downgrade_threshold`, `ladder_up_successes`, `channel_guard_ms`) are all config-tunable. Defaulting ON at 2000 ms with no field mitigation is a bit of a one-way door. I know the project avoids proliferating ARQ knobs, so this may well be intentional — just flagging the asymmetry.

3. **Doc nit:** the body says "0–2000 ms"; `x % jitter_ms` yields [0, 1999].

4. **Note for whoever merges this into #160.** That branch has its own retry-timing work: `21c01f3` re-anchors the ACCEPT retry to a heard CALL using `ARQ_CHANNEL_GUARD_MS`. That site must stay un-jittered — it is a channel guard, not a retry interval — and your rule already gets that right, since it only wraps `retry_interval_s`-based deadlines. Just worth not "fixing" during conflict resolution.

Good, well-scoped change on a real field bug, and the salt design is the part I would have got wrong.

pedromessetti:
@rafael2k thanks — that reading is right, and the `ack_timeout_s` exclusion was the gap. I've jittered it too.

Changes pushed:

1. **Jitter the DATA retransmit** (`arq_fsm.c` `ARQ_DFLOW_DATA_TX → WAIT_ACK`). Your safety argument is exactly correct: `ack_timeout_s` is a lower bound, so a bounded *positive* jitter only ever waits longer and can't violate the margin. It is now the load-bearing timer for the countdown scenario. Added `test_wait_ack_deadline_includes_bounded_jitter` pinning the bounded range on that deadline.

2. **Robust divergence tests.** Replaced the two one-pair assertions with 100-instant samples requiring several distinct outcomes, so a single hash collision (or a future change to the constants/seed timestamps) can't produce a baffling one-off failure.

3. **Doc nit fixed** — PR body now says `[0, jitter_ms)` / 0–1999 ms.

On the smaller points:

- **Runtime off-switch** — intentional for now: `arq_retry_jitter_ms` is an `_Atomic int`, so it is already settable to 0 at runtime (there's a test exercising the disabled path), but I kept it out of the `.ini`/TCP knobs to avoid proliferating ARQ config. Happy to add plumbing if you'd rather have field mitigation.

- **#160 ACCEPT re-anchor** — agreed; only `retry_interval_s`-based deadlines are wrapped, so that channel-guard site stays un-jittered. I'll leave it untouched during any merge/conflict resolution.

- **Latency note** — added to the PR body.

Full suite still green (23 binaries, 0 failures; `test_arq_fsm` now 37, `test_arq_protocol` 31).

rafael2k:
Re-measured the updated branch, and then @rafael2k raised a design objection I think is right and want to put on the record before this lands.

## The measurement first

Built `6b0bfa5` and ran it against trunk and against the control-only version, same harness, only the binary varies:

| cell | trunk | control-only jitter | +DATA jitter |
|---|---|---|---|
| 1 KB @ 5.2 dB | 101 s | 101 s | **1 no-connect**, then 101 / 102 / 100 s |
| 102 B @ −4.8 dB | 226 s | 247 s | 225 s |

Unit suite: 271 assertions, 0 failures. The fringe cell is now *better* than the control-only version (225 s vs 247 s), so jittering the DATA retransmit did not cost what I feared.

The one thing worth flagging: one run in four failed to connect on a clean 5.2 dB link, and did not reproduce in three repeats. Trunk and the control-only build never failed there in any run. I am not claiming that as a defect on n=4 — but it is exactly the failure shape randomness produces, and exactly the kind that is miserable to chase later.

## The design objection: why randomness on the DATA plane at all?

Random backoff is the right tool when contenders are **many, anonymous, and cannot coordinate** — ALOHA, CSMA, Ethernet. None of that holds here. Mercury's ARQ is **point-to-point between two stations that have exchanged callsigns**: `sess->local_call` and `sess->remote_call` are both populated at CALL/ACCEPT. Both ends hold the same pair of identities, which is shared deterministic state. When you have identity, you do not need luck to break symmetry — you can just *decide*.

A deterministic per-node offset — the "color code" idea — gets the same result strictly better:

```c
/* Both ends compute the SAME comparison from the SAME pair of callsigns,
 * so they always disagree about who defers.  No draw, no tail. */
int rank = (strcmp(sess->local_call, sess->remote_call) > 0) ? 1 : 0;
deadline = now + base + rank * ARQ_RETRY_STAGGER_MS;
```

Against random jitter this is better on four counts:

1. **Reproducible.** A bad run replays exactly. The no-connect above might take twenty runs to see again — with a deterministic stagger there is nothing to reproduce.
2. **No tail latency.** One station pays 0 and the other pays a fixed, known Δ. Random jitter costs an expected Δ/2 on *every* retry for *both* stations, plus the tail.
3. **Testable as equality.** This PR had to relax `test_calling_reanchors_retry_on_tx_complete` from exact equality to a bounded range, and replace two divergence assertions with 100-instant sampling. A deterministic stagger lets all of those go back to asserting exact deadlines.
4. **Bounded worst case,** rather than "usually fine".

The machinery is already here — `arq_protocol_node_salt()` mixes the callsign; it just needs to yield a *stable rank from the pair* instead of feeding a per-retry hash of the clock.

**The one case that genuinely needs randomness** is the degenerate one the deterministic rule cannot separate: both stations configured with the same callsign. Falling back to jitter only when `strcmp(local, remote) == 0` covers it, and that is a misconfiguration rather than the normal path.

## What I would keep

The `ack_timeout_s` site was the right find and should stay jittered/staggered whichever mechanism wins — it is the timer that actually locks in #217, and the lower-bound argument holds for a fixed offset just as it does for a random one. Nothing about this objection undoes that part of the fix.

pedromessetti:
@rafael2k agreed — point-to-point with exchanged callsigns shouldn't need luck to break symmetry. Replaced the random jitter with the deterministic stagger.

Implemented:

- `arq_protocol_retry_rank(local, remote)` returns `0`/`1` from `strcmp(local, remote) > 0`, and `-1` for the degenerate pair.
- `deadline = now + base + rank * retry_stagger_ms`. The offset accumulates across retransmits (each deadline re-anchors at PTT-OFF), so the two stations drift apart and one soon transmits while the other listens.
- Fallback to a random draw only when `strcmp == 0` or a callsign is missing.
- The `ack_timeout_s` site stays staggered (as you said — the timer that actually locks in #217).
- Tests restored to exact-equality assertions; `test_arq_protocol` gains antisymmetry + degenerate-pair tests.

One correction to the sketch: `sess->local_call` is *not* populated on the caller side — it only carries the dialed SSID on the callee (`empty => fall back to the primary`, per `arq_fsm.h`). Using it directly would make the caller always compare `""` vs the remote and rank `0`, breaking antisymmetry when the caller's callsign sorts after the peer's. So the FSM uses:

```c
const char *local = sess->local_call[0] ? sess->local_call : my_call;  // primary via arq_conn_get_calls
```

That keeps the comparison over the same exchanged pair on both ends (caller primary ↔ dialed SSID), so `strcmp` stays antisymmetric.

Renamed the knob to `arq:retry_stagger_ms` (default 2000, 0..5000, 0 disables). Full suite green (23 binaries, 0 failures).

rafael2k:
Re-reviewed at `810c723`, and merged `mercuryv2` in to resolve a conflict that #215 had just created — details at the bottom.

## The refactor is right, and it is verified

You took the deterministic-stagger argument and implemented it cleanly. The core property is that `strcmp` is antisymmetric, so the two peers of a session — comparing the *same* pair of exchanged callsigns — can never agree on who defers. I checked that directly rather than by eye, evaluating each session from both ends:

```
  PU2UIT-2   <-> PU2UIT-3    A: 0 B: 1  exactly one defers
  A          <-> ZZZZ        A: 0 B: 1  exactly one defers
  PU2UIT-10  <-> PU2UIT-9    A: 0 B: 1  exactly one defers
  pu2uit     <-> PU2UIT      A: 1 B: 0  exactly one defers
  PU2UIT     <-> PU2UIT      A:-1 B:-1  degenerate -> random fallback (both)
  PU2UIT     <->             A:-1 B:-1  degenerate -> random fallback (both)
```

The degenerate rows are the important ones: both ends independently reach `-1`, so they *both* fall back to a random draw. A design where one end went deterministic and the other random would be worse than either — this gets it right.

Two details I specifically went looking for and found correct:

- **Case and SSID differences are safe.** I half-expected a hazard where one side normalises a callsign and the other does not, making the two `strcmp` calls compare different byte strings and return the same sign. There is no `toupper()` anywhere in `datalink_arq/` or `common/`, so callsigns travel and are stored verbatim — both peers compare the identical pair of byte strings, whatever their case. The row above (`pu2uit` vs `PU2UIT`) still resolves correctly.
- **The `arq_fsm.c` fallback is sound.** `sess->local_call[0] ? sess->local_call : my_call` degrades to `rank == -1` (random) when identity is not yet established, rather than to a fixed rank both ends might share.

Keeping the `ack_timeout_s` site staggered is the right call, and the naming change throughout (`jitter` → `stagger`) makes the mechanism self-describing.

## Testing

Built at `810c723` and ran everything locally:

- Unit suite: **272 assertions, 0 failures**
- Integration suite: `ok mercuryintegration 274.054s`
- All 9 CI checks green, including TSan and ASan+UBSan

The single failed connect I saw in the random version did not recur — which is the point: with the draw gone there is no longer a rare instant that produces it.

## The conflict I resolved

#215 merged an hour ago and added PTT round-trip fields to `tests/common/test_cfg_utils.c` — the same config round-trip test where this PR renames `retry_jitter_ms` to `retry_stagger_ms`. Two hunks, in the `w` population block and the matching `r` assertion block.

These are independent additions to the same test rather than competing edits, so I kept **both** sides — trunk's five PTT fields and this PR's `retry_stagger_ms = 0` off-switch line, with the assertion to match. I checked them per-block rather than taking both sides blindly, since a union resolution in C is normally how you get something that compiles and is quietly wrong; here both blocks are additive field assignments touching no shared field.

Rebuilt after the merge: **284 assertions, 0 failures**.

Merging. Thanks for turning this around so quickly, and for taking the design argument on its merits.

PR #222, opened 2026-08-26
Fix audio capture on Mac OS

MobiusHorizons:
Mac OS requires the app to specify the `NSMicrophoneUsageDescription`  property in the Info.plist in order to request permission from the user for accessing the audio capture device. This change uses a builtin utility on MacOS (`PlistBuddy`) to add the missing property.


rafael2k:
Thanks for this — the diagnosis is right and the fix is the correct shape. macOS will hand a sandboxed/hardened app silence rather than an error when `NSMicrophoneUsageDescription` is missing, which is about the worst failure mode for a modem: it looks like a dead band, not a permissions problem. Worth fixing.

Three things before merge, then some context on the red check.

### The failing macOS check is not yours

`Build macOS arm64` fails with `hamlib/rig.h file not found`, which is a pre-existing break, not something this patch causes. The branch is based on `dc26a95` (12 Aug) and is **124 commits behind** `mercuryv2`. The cause was `UNAME_S` being assigned only inside the `HAVE_HERMES_SHM` probe in `radio_io/Makefile` — a block the top-level Makefile skips, so `UNAME_S` was empty, the `Darwin` arm of the hamlib block was never taken, and macOS silently fell through to pkg-config instead of using the vendored `radio_io/hamlib-macos`. That is fixed on `mercuryv2` now, and the macOS job passes there. **A rebase should turn the check green** with no change to your commit.

### But the green checks don't cover you either

Nothing in CI runs `fyne-ui-macos`, `fyne-ui-macos-universal`, or either `-dmg` target — the macOS job is build-only (`make`), by design. So this line is exercised only when a human runs the packaging target on a Mac. My macOS VM is currently down so I could not verify the behaviour below on a device; please confirm on yours.

### 1. `Add` is not idempotent — it will break the second run

`PlistBuddy -c "Add ..."` exits non-zero with `Entry Already Exists` if the key is present. If `fyne package` ever leaves an existing `Info.plist` in place (rather than regenerating it), the second `make fyne-ui-macos` in the same tree aborts the build. Running a packaging target twice is the normal case, and this is the release path. The usual robust form:

```make
	$(PLIST_BUDDY) -c "Set :NSMicrophoneUsageDescription $(MACOS_MIC_USAGE)" $(MACOS_PLIST) 2>/dev/null || \
	$(PLIST_BUDDY) -c "Add :NSMicrophoneUsageDescription string $(MACOS_MIC_USAGE)" $(MACOS_PLIST)
```

### 2. The string is duplicated verbatim in two targets

`fyne-ui-macos` and `fyne-ui-macos-universal` carry the same 70-character sentence. A future wording change will update one and drift the other. Define it once near `MACOS_APP_ID`:

```make
MACOS_PLIST      = $(FYNE_UI_DIR)/$(MACOS_APP_NAME).app/Contents/Info.plist
MACOS_MIC_USAGE ?= Mercury needs access to the radio audio to listen for data signals
```

### 3. Please confirm the quoting

`PlistBuddy -c "Add :Key string 'text with spaces'"` is the common idiom and I believe PlistBuddy's own parser strips the single quotes — but I could not test it here. Since the string is user-visible (it is what the permission dialog shows the operator), it is worth one check that the quotes are not stored literally:

```
/usr/libexec/PlistBuddy -c "Print :NSMicrophoneUsageDescription" fyne-ui/Mercury.app/Contents/Info.plist
```

Should print the sentence with no surrounding `'`.

### One thing you got right that is easy to lose

Both insertions land **after** `fyne package` and **before** any `$(call macos_sign,...)` (Makefile:426/446/450 in `fyne-ui-macos-universal-dmg`). That ordering matters: editing `Info.plist` after `rcodesign` runs would invalidate the bundle seal and the signed build would fail Gatekeeper. Please keep the line where it is if you restructure.

With the rebase and points 1–2 addressed, this is good to merge from my side.

PR #232, opened 2026-09-01
watterson: seed the fading channel, from a private generator

rafael2k:
## The bug

`common/watterson.c` draws its Rayleigh taps and AWGN from `rand()`, and **nothing on that path ever called `srand()`**. libc then behaves exactly as if seeded with 1, so every run produces the *identical* fading realisation.

This is worse than a missing feature, because it looks like agreement. Four "independent" runs of a connect sweep at SNR3k −10.8 dB all decoded at +27.107 s with `snr=-5.78`, and were reported as 4/4. It was **one sample printed four times**. Every fading A/B measured through this tool has the defect.

Demonstrated directly on the pre-fix code — two runs of `watterson_test --moderate` are byte-identical:

```
6bdf9e85435fa3c92aa53eb8154897f6  trunk_run1.raw
6bdf9e85435fa3c92aa53eb8154897f6  trunk_run2.raw
```

## Why a private generator, not just an `srand()` call

Ported from mercuryv1's `source/common/os_interop.cc`, reduced to the single glibc TYPE_3 degree Mercury needs and made reentrant (cursors are indices, so the struct is trivially copyable).

- **Reproducible across platforms.** `rand()` is a different generator on glibc, musl, macOS and MinGW — the same seed reproduces a *different* channel on each. A fading measurement nobody else can replay cannot be checked.
- **No shared global state.** `rand()` has one hidden sequence, so any other caller silently reorders the draws the channel sees. State now lives in `watterson_t`, per instance.
- **Hard to leave unseeded.** `watterson_init()` seeds from `MERCURY_WATTERSON_SEED`, else clock XOR pid, and *records* the seed; `watterson_test` prints it, so an unpinned run stays replayable after the fact.

## This is a verified no-behaviour-change

Not merely a plausible one. The uniform draw is deliberately spelled to match the expression it replaced — `(float)rand() / RAND_MAX`, a *float* divide by `(float)2147483647` — and `M_PI` stays a `double` in the Box-Muller step. The Doppler IIR's poles sit within ~1e-4 of the unit circle, so a last-bit difference in the driving noise gives a completely different fade a few thousand samples later.

With the seed pinned to 1 (what unseeded `rand()` effectively was), the new channel is **byte-identical to the old**:

```
  --moderate No=-6   IDENTICAL      --poor No=-6   IDENTICAL      --good No=-6   IDENTICAL
  --moderate No=0    IDENTICAL      --poor No=0    IDENTICAL      --good No=0    IDENTICAL
  --moderate No=-20  IDENTICAL      --poor No=-20  IDENTICAL      --good No=-20  IDENTICAL
```

And afterwards: two unpinned runs differ; two runs at `MERCURY_WATTERSON_SEED=777` are identical.

`utils/chanutil.c` seeded libc rather than the model; it now calls `watterson_seed()`, which is what it always meant.

## Tests

- `tests/common/test_mercury_prng.c` — pins the generator against a **hardcoded** glibc reference stream (hardcoded, not read from the host's `random()`, which would only prove we match *this* libc — the very dependency being removed); proves two instances cannot perturb each other; asserts the uniform draw is bit-exact with the old expression.
- `test_watterson.c` — golden FNV-1a of the seed-1 realisation. The fade depends on the **order and count** of draws, not just the generator, so a refactor that merely moves a draw would otherwise invalidate every pinned measurement without failing anything. This guard already earned its keep: it caught a bug in this PR's own first draft, where reporting the seed *re-seeded* the model after `watterson_add_path()` had already drawn.

Full suite green; `make`, `make -C utils`, `make -C tests` all clean.

## Consequence

Any fading result in this repo's history that was measured through this tool is one realisation, not a sample. They need re-measuring before being relied on — I've already retracted one of my own on that basis.

### DIFF
--- a/common/mercury_prng.c
+++ b/common/mercury_prng.c
@@ -1,0 +1,132 @@
+/* HERMES Modem — reproducible pseudo-random source
+ *
+ * Copyright (C) 2026 Rhizomatica
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * See mercury_prng.h for why this is not just rand().
+ */
+
+#include "mercury_prng.h"
+
+#include <math.h>
+#include <stdlib.h>
+#include <time.h>
+#ifdef _WIN32
+#include <process.h>
+#define MERCURY_GETPID() _getpid()
+#else
+#include <unistd.h>
+#define MERCURY_GETPID() getpid()
+#endif
+
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
+
+/* Separation between the front and rear cursors for TYPE_3 (x**31 + x**3 + 1). */
+#define SEP 3
+
+void mercury_prng_seed(mercury_prng_t *p, unsigned int seed)
+{
+    int32_t word;
+    int     i;
+
+    if (!p)
+        return;
+
+    /* glibc maps 0 to 1: seeding the additive generator with an all-zero
+     * state would make it emit zeros forever. */
+    if (seed == 0)
+        seed = 1;
+
+    p->state[0] = (int32_t)seed;
+    word = (int32_t)seed;
+
+    /* state[i] = (16807 * state[i-1]) % 2147483647, computed by Schrage's
+     * method so the intermediate product never overflows 31 bits. */
+    for (i = 1; i < MERCURY_PRNG_DEG; ++i)
+    {
+        long hi = word / 127773;
+        long lo = word % 127773;
+        word = (int32_t)(16807 * lo - 2836 * hi);
+        if (word < 0)
+            word += 2147483647;
+        p->state[i] = word;
+    }
+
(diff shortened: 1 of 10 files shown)


rafael2k:
Follow-on: seeding `watterson_init()` was necessary but not sufficient for the integration harness.

The channel bridge runs a separate `watterson_test` process per direction, and both inherited the same `MERCURY_WATTERSON_SEED` — so forward and reverse faded in lockstep. A deep fade blocked an outgoing CALL and the returning ACCEPT at the same instant, and asymmetric-link effects were invisible by construction. Fixed by salting the seed per direction (`22f4d34` on `mfsk-margin`).

Worth knowing for anyone reading old fading measurements: results taken before both fixes are one realisation, not a sample, and A/B comparisons across branches were not necessarily on the same channels.

PR #233, opened 2026-09-01
fyne-ui: make the embedded client an explicit, interlocked choice

rafael2k:
## The situation

`mercury-ui` embeds the engine, but its chat client is an **ordinary TCP client of our own TNC ports** — 127.0.0.1 control/data/broadcast, exactly like uucp's `uuport`, VarAC or RNS (`gui_interface/mercury-client/client/client.go:81-84`). It is not privileged and **cannot hold the ports hostage**.

The hazard runs the other way. The TNC keeps exactly one control and one data client (single fds, not arrays), and a new `accept()` **evicts the incumbent**:

```c
/* data_interfaces/tcp_interfaces.c:799 */
if (ctl_client >= 0)
    close_ctl_client(&ctl_client, &data_client, true);
```

That `true` is `notify_arq`, which submits `ARQ_CMD_CLIENT_DISCONNECT` → `ARQ_EV_APP_DISCONNECT` (`arq.c:565`) — **tearing down any live ARQ session**.

So today, pressing *Connect modem* during a uucp transfer kills the transfer, and the operator's only clue is that the transfer failed. The reverse is just as quiet: starting uucp takes the ports away from the chat window, whose reader goroutines simply `return`, leaving it showing "connected" with buttons that no longer do anything.

## Three changes

**1. A master switch.** An "Embedded client" checkbox in the top bar gates the Launch button, persisted in preferences — so a station that exists to serve uucp can turn the chat client off once and have it stay off. Switching it off also closes an open window, because otherwise "off" would only mean *cannot be opened again* while a connected client kept holding the ports. **Defaults ON**, so nothing changes for anyone who hasn't asked for it.

**2. An interlock on connect.** If another client already holds the ports, refuse and say so rather than silently evicting it:

> Another client is attached to the TNC ports AND an ARQ session is live.
>
> Connecting here would disconnect it and abort any transfer in progress. Stop the other client first.

Skipped when the attached client is *us* — then there's nobody else to evict, and reconnecting must stay possible. The decision is split into `embeddedClientMayConnect()` so it's testable without a running engine.

**3. Link-down detection.** `ModemClient` now records that the modem hung up on us, instead of letting `IsConnected()` answer true off conn objects that are merely non-nil. The chat window watches for it and returns to the disconnected state with an explanation.

## Tests

`TestEmbeddedClientMayConnect` covers all six states, including the two that must still be *allowed* (the attached client is us, with and without a live session) — a too-eager interlock that locks the operator out of their own client would be its own bug.

## Verification

`go build` / `vet` / `test` clean for both Go modules, `make fyne-ui` links, C suite unchanged.

**Not verified:** the UI behaviour itself hasn't been exercised on a running station — it wants a click-through with uucp attached.

### DIFF
--- a/common/os_interop.h
+++ b/common/os_interop.h
@@ -22,6 +22,7 @@
 
 /* Socket compatibility */
 #define SOCK_CLOSE(fd) closesocket(fd)
+#define SOCK_SHUTDOWN(fd) shutdown(fd, SD_BOTH)
 #define SOCK_IOCTL(fd, cmd, argp) ioctlsocket((SOCKET)(fd), (long)(cmd), (u_long*)(argp))
 #ifndef MSG_NOSIGNAL
 #define MSG_NOSIGNAL 0
@@ -153,6 +154,7 @@ int COND_SIGNAL(HANDLE *mqh_wait);
 
 /* Socket compatibility */
 #define SOCK_CLOSE(fd) close(fd)
+#define SOCK_SHUTDOWN(fd) shutdown(fd, SHUT_RDWR)
 #define SOCK_IOCTL(fd, cmd, argp) ioctl((fd), (unsigned long)(cmd), (argp))
 
 static inline int sock_set_nonblocking(int fd)
--- a/data_interfaces/tcp_interfaces.c
+++ b/data_interfaces/tcp_interfaces.c
@@ -1469,9 +1469,39 @@ void *tcp_server_thread(void *port_ptr)
         if (!send_started)
             bcast_client_done = true;
 
-        /* Stop and join the sole RX-buffer reader before draining its ring.
-         * Clearing before join can empty the ring after the sender's size/
-         * done checks but before read_buffer(), stranding it in a wait. */
+        /* recv_thread normally notices a clean FIN (recv()==0) and exits on
+         * its own, letting the join below return.  Do not trust that alone:
+         * if the client half-drops the link without a FIN -- or a non-blocking
+         * read races the close -- recv_thread spins on EAGAIN forever and this
+         * thread never returns to accept(), so the listen backlog fills and the
+         * NEXT client's connect() hangs.  Watch the socket directly and, on
+         * hangup/error, shut it down so recv() unblocks.  shutdown() keeps the
+         * descriptor owned by this thread until the join, avoiding the
+         * cross-thread close race. */
+        while (!shutdown_)
+        {
+            struct pollfd cfd = { .fd = client_socket, .events = POLLIN | POLLERR | POLLHUP };
+            int r = poll(&cfd, 1, 100);
+            if (shutdown_)
+                break;
+            if (r < 0 && sock_errno() == SOCK_EINTR)
+                continue;
+            if (r <= 0)
+                continue;
+            if (cfd.revents & (POLLERR | POLLHUP | POLLNVAL))
+                break;
+            if (cfd.revents & POLLIN)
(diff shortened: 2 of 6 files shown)


pedromessetti:
## Finding 1 (HIGH) — the interlock is bypassed exactly when it matters

`watchLink` detects the eviction and flips the UI back to "disconnected", but it does **not** clear `cw.mc`:

```go
// mercury_chat_window.go:720-723
cw.logMsg("Disconnected by the modem: another client has taken the TNC ports.")
cw.setTCP(false)
cw.setARQ(false)
return
```

`cw.mc` is left pointing at the now-dead client, and the interlock's "already ours" test is `cw.mc != nil`:

```go
// mercury_chat_window.go:440
if ok, why := embeddedClientMayConnect(cw.mc != nil, currentTelemetry()); !ok {
```

So the sequence this PR is trying to prevent still happens:

1. Embedded client is connected (`cw.mc != nil`, holds ports).
2. uucp attaches → TNC evicts us → `linkDown` → `watchLink` returns to disconnected, `connectBtn` re-enabled, but `cw.mc` still non-nil.
3. Operator clicks "Connect modem" → `alreadyOurs == true` → interlock says yes → we connect and evict uucp back.

That is the very eviction the interlock exists to stop, and it fires in the one flow where a foreign client is present.

**Fix:** `watchLink` should mirror `onDisconnect` — clear `cw.mc` (and close `cw.done`) on the UI thread so the next Connect is judged with `alreadyOurs == false`. This also stops the four `forward*` goroutines from leaking until the next connect/close.

pedromessetti:
## Finding 2 (MEDIUM) — misleading "another client took the ports" on a normal disconnect

```go
// modem.go:468-479
line, err := reader.ReadString('\r')
if err != nil {
    if err != io.EOF {
        mc.LogCh <- fmt.Sprintf("ARQ Control read error: %v", err)
    }
    if !mc.linkDown.Swap(true) {
        mc.LogCh <- "ARQ control link closed by the modem " +
            "(another client may have taken the TNC ports)."
    }
    return
}
```

`Disconnect()` closes `ARQControlConn` locally, which wakes this blocked read and lands in the same error branch. The code never checks the `quit` channel, so a deliberate local disconnect is also reported as "another client may have taken the TNC ports." It is racy (the message only shows if `forwardLog` reads it before `done` closes), but it is a wrong diagnosis.

**Fix:** before attributing EOF to the modem, check whether we asked to go away — `select { case <-quit: return; default: }` — and (to close the race) close `quit` in `Disconnect()` before closing the conns, so the flag is already observable when the read wakes.

pedromessetti:
## Finding 3 (LOW) — `watchLink` re-enables `arqConnect` while disconnected

`watchLink` calls `setTCP(false)` then `setARQ(false)` (lines 721-722), and `setARQ(false)` unconditionally does `cw.arqConnect.Enable()` (line 406). Net result: `arqConnect` ends up enabled with no connection. It is harmless — `onARQConnect` guards on `mc.IsConnected()` — but the button state is wrong.

This same ordering already exists in `onDisconnect` (lines 509-510), so it is a pre-existing quirk this PR copies.

**Fix:** gate the `arqConnect.Enable()` on the modem still being connected, matching the existing `sendBcastWrap`/`sendCQ` gating:

```go
if cw.mc != nil && cw.mc.IsConnected() {
    cw.arqConnect.Enable()
    cw.sendBcastWrap.Enable()
    ...
}
```

PR #234, opened 2026-09-01
modem/mfsk: Chase-combine MFSK bursts that individually fail to decode

rafael2k:
## The idea

The MFSK floor's retry sequence **already is gapped repetition**: a burst that fails is re-sent seconds later through an independently faded channel. But each copy was decoded alone and thrown away, so three failed attempts stayed three failures — even when their errors fell in different places and the copies *together* carried the codeword.

This accumulates the demodulated LLRs across attempts and decodes the sum.

## Measured

`tests/modem/test_mfsk_harq.c`, at an SNR chosen so a single copy never makes it:

```
single copy :   0/200 decoded
3 combined  : 200/200 decoded
```

## Bounded, deliberately

An unbounded accumulator is not diversity, it's a trap. Copies from *different* transmissions are different codewords, and summing those is noise that would actively **prevent** decoding — worse, a stale partial sum could poison an unrelated later burst.

So the accumulator holds at most `MFSK_HARQ_MAX_COPIES` (4) over at most `MFSK_HARQ_WINDOW_MS` (90 s), and on overflow **discards and restarts** from the current copy rather than blending old and new.

The third test case is the one that matters for safety — feed copies of two different codewords and require that nothing decodes:

```
mixed pair  :   0/200 converged, 0 matched a real codeword
```

Not a false decode, and not merely a decoder that fails to converge. Neither is allowed to look like a delivery.

## Scope

Split out of `deep-connect` on purpose: **this is the part with evidence behind it.** The CALL-escalation work it sat next to measured *identical to baseline, seed for seed*, once the fading channel was actually seeded (#232) — so it is not proposed here. The original `0/4 → 4/4` claim for that work was an artifact of the frozen channel and is retracted.

Three files, no unrelated changes.

## Verification

- `make -C tests test` — full suite green on the rebased branch
- `make` clean
- Base branch was merged up to trunk first (`62fdd8f` + `cbb2438`), which required repairing a build break: #231's `arq_get_peer_snr_x10()` reads session fields this branch dropped with OLLA. Peer SNR is now kept as display-only telemetry, explicitly not fed back into mode selection.

**Not verified on air.** This is a decoder-level change proven by unit test; its value on a real fading link is untested.

### DIFF
--- a/modem/modem_mfsk.c
+++ b/modem/modem_mfsk.c
@@ -9,6 +9,13 @@
 #include "mfsk_ofdm.h"
 #include "mfsk_sync.h"
 #include "mfsk_ldpc.h"
+
+/* How much may be Chase-combined before the accumulator is abandoned.  Sized
+ * from the MFSK CALL retry cadence (~18 s apart, at most a handful of tries):
+ * long enough to hold one genuine retry sequence together, short enough that an
+ * abandoned sequence cannot reach the next caller. */
+#define MFSK_HARQ_MAX_COPIES 4
+#define MFSK_HARQ_WINDOW_MS  90000ULL
 #include "freedv_api.h"   /* freedv_gen_crc16 — codec-independent CRC16 */
 
 #include <complex.h>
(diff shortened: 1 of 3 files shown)


rafael2k:
Base branch is now current with trunk (`62fdd8f` + `cbb2438`), so this rebases cleanly.

Independent of #235: this is the decoder-level change (proven by unit test), that one is the CALL carrier change (proven by paired sweep). They touch different files and can go in either order.

Issue #237, opened 2026-09-02
Broadcast TX ingest: raw-vs-wrap detection is fragile


rafael2k:
`bcast_process_client_frame()` decides whether a client's `CMD_DATA` frame is already a raw modem frame or needs Mercury's header + 2-byte length prefix injected. The test is a heuristic on the frame's own first byte:

```c
is_raw_modem_frame = (pt == BROADCAST_CONTROL || pt == BROADCAST_DATA) &&
                     (frame_header_extension(decoded_frame[0]) == 0 ||
                      (size_t)frame_len == frame_size);
```

The `frame_len == frame_size` clause was added in 70ff1ef to stop hermes-broadcast's joint RaptorQ frames being truncated (they fill the frame exactly and carry a session id in the extension field, so the original `extension == 0` test rejected them and wrapping cut 1180 to 1177 — silently, and the far side dropped all of them).

That fixes the symptom but the underlying design is still guesswork: we are inferring the sender's intent from bytes that a different protocol is free to use for its own purposes. Two known sharp edges:

- `BCAST_EXT_LEN_PREFIX` is `0x01`, so any odd session id sets it. A short hermes-broadcast frame (not full frame_size) would still be misread.
- A client sending genuinely unformatted data that happens to begin with a byte whose top 3 bits are 3 or 4 gets passed through unwrapped.

Better would be for the client to state its framing explicitly rather than have Mercury guess — e.g. a distinct KISS command value for "already a modem frame", or a capability exchanged when the broadcast client connects. That is a wire-visible change affecting VARA clients, hermes-broadcast and RNS, so it wants designing rather than patching.

Not urgent: the current logic works for every case we test (VARA/chat short frames still wrap, full-size broadcast frames pass through).

rafael2k:
Pinned the actual behaviour in tests (`test_bcast_tx_full_size_chat_is_still_wrapped`, `test_bcast_tx_full_size_cmd_data_passes_through_raw`). The answer differs by sender, which narrows this issue usefully.

**Chat cannot be misclassified.** `needs_wrap` is unconditionally true for `CMD_AX25` and `CMD_AX25CALLSIGN`:

```c
bool needs_wrap = (kiss_cmd == CMD_AX25 || kiss_cmd == CMD_AX25CALLSIGN) ||
                  (kiss_cmd == CMD_DATA && !is_raw_modem_frame);
```

[...]

PR #238, opened 2026-09-02
Broadcast file transfer (RaptorQ) with a Fyne UI panel

rafael2k:
Mercury could carry broadcast chat but not files. This adds RaptorQ file broadcast, wire-compatible with **hermes-broadcast's `broadcast_daemon`**, plus send and receive in the Fyne UI.

## Verified end to end

Between two real Mercury modems over the `ch` channel — file recovered **byte-identical, under its original name**:

```
mode  9 (DATAC17)  1 kB  clean           2 s
mode  9 (DATAC17)  5 kB  clean          31 s
mode  0 (DATAC1)   5 kB  clean          50 s
mode  9 (DATAC17)  1 kB  SNR3k 10.2 dB   2 s
```

`utils/bcast_ota_test.sh` is the harness. At SNR3k −0.8 dB mode 9 fails, which is DATAC17's own threshold.

## Design notes

**Joint self-describing frames.** Every frame carries header + reduced OTI + tag + symbol, as `broadcast_daemon` does — 8 bytes more per frame than the older split format, but a receiver with no return path starts decoding on the *first frame it hears* instead of waiting for a periodic config frame.

**One source block.** nanorq defaults to Z=16; RaptorQ codes each block independently, so that costs the fountain's overhead 16 times (measured 5–17% more frames) **and** has a failure mode — loss periodic in the carousel starves one block completely (0 symbols of 15000 dropped, never decoding). Z=1 removes both; falls back to the smallest legal Z where a single block cannot fit.

**The filename travels in a bundle** (mercury-connector's `spool.c` layout). An incoming bundle is untrusted: the parser refuses path separators, NULs, `.`/`..`, a size field disagreeing with the buffer, and a missing terminator.

**One socket.** The UI's file transfer rides the chat client's broadcast connection — Mercury's broadcast port accepts exactly one client. A frame filter gives the file receiver first refusal; everything else goes to chat.

**Mode is reported, not offered.** `-m` is fixed at engine start and there is no runtime switch, so the panel shows what to set the far station to, in words rather than an index:

```
Mode: DATAC17 - fastest, needs a strong signal
1407 bit/s, 2062 Hz bandwidth, 1180 bytes per frame
The receiving station must use the same mode (mercury -m 9).
```

Bitrate and bandwidth are computed from the **running modem**, not a table (bandwidth via a new `freedv_get_modem_bandwidth_hz()`), so they cannot drift from what is on the air. Cross-checked against codec2's published figures.

Progress is frames and cycles, never a percentage — broadcast has no return path, so a percentage would be a guess about someone else's decoder.

## Included fix

`tcp-bcast` was **silently truncating** full-size broadcast frames from a client (1180 → 1177), because the raw-vs-wrap test demanded a zero extension field and these carry a session id there. Every frame was then dropped by the far side. See #237 for why that detection is still fragile and wants redesigning.

## Portability

nanorq is vendored; Mercury shells out to nothing at runtime. Two fixes Windows needed: the mmap ioctx is compiled out on `_WIN32`, and nanorq's own `include/io.h` **shadows the system `<io.h>`** so mingw's `ftruncate` inline could not find `_chsize`. Builds verified for native, mingw-w64 and aarch64.

## Tests

15 C tests (carousel through 20–30% loss on modes 0/8/10, bundle round-trip and hostile-input refusals, periodic-loss regression, cycle budget) and Go tests covering the CGo marshalling by driving send into receive through the real carousel while dropping every third frame.

**Not verified:** the UI has not been clicked through on a running station.


rafael2k:
Feature complete and documented — [docs/BROADCAST-FILE.md](docs/BROADCAST-FILE.md), linked from the README.

The Windows failure was mine and is fixed in `56228a6`: `bcast_file_tool` uses BSD sockets and is now Linux/macOS-only (the Windows UI runs the same code in-process over CGo), and nanorq's vendored `io.h` was shadowing the system `<io.h>` on mingw — it is vendored as `nanorq_io.h` now, so the earlier `_chsize` workaround is gone rather than relocated. All 9 checks green.

## Ready for OTA

Bench results (`utils/bcast_ota_test.sh`, two real modems with `ch` as the air, byte-identical recovery with the original filename):

| mode | file | channel | time |
|---|---|---|---|
| 9 DATAC17 | 1 kB | clean | 2 s |
| 9 DATAC17 | 5 kB | clean | 31 s |
| 0 DATAC1 | 5 kB | clean | 50 s |
| 1 DATAC3 (default) | 1 kB | clean | 32 s |
| 9 DATAC17 | 1 kB | SNR3k 10.2 dB | 2 s |

### Suggested OTA procedure (estacao2 / estacao3)

Both stations must run the **same mode** — there is no negotiation on the broadcast plane.

```
# both stations, e.g. the default DATAC3
mercury -v -S -m 1

# receiver
utils/bcast_file_tool recv /root/bcast_rx -m 1

# sender
utils/bcast_file_tool send /root/test_file -m 1 -c 0
```

`-c 0` carousels until interrupted; stop the sender once the receiver reports the file. The receiver writes it under the sender's own filename, which is the thing to confirm on air.

Worth trying at DATAC3 first (robust, 316 bit/s) and then DATAC1 or DATAC17 if the path supports it. A 5 kB file at DATAC3 will take a few minutes of continuous carousel.

**What OTA would actually test that the bench cannot:** real fading rather than `ch`'s AWGN, real PTT timing around the carousel, and whether a receiver joining mid-transmission recovers — the bench always starts the receiver first.

The one part still unexercised anywhere is the UI widget wiring itself (dialogs, toggle, folder chooser); everything beneath it is covered by tests and the harness.

rafael2k:
Rebased onto trunk now that #239 has landed, as offered there.

The two changes touched the same `libmercury_core_w64.a` rule and are both needed, so the merged recipe carries both:

```make
$(MINGW_AR) rcs $@ $(MERCURY_CORE_OBJS_W64) $(HIDAPI_W64_OBJ) $(BCAST_FILE_OBJS_W64) \
    $(FYNE_UI_DIR)/engine/mercury_bridge_w64.o
```

Verified the archive is actually complete rather than just building — every `hid_*` referenced by `cm108_ptt.o` is defined by `hid.o` within the archive, and the RaptorQ objects are present:

```
hid.o present:    1
raptorq objects:  5
all hid_* referenced are also defined within the archive
```

(`nm` on an archive lists per-object undefineds, so a raw `U hid_` count is not a useful check here — `cm108_ptt.o` will always show them.)

From clean: native build, C suite, Go tests and `make windows` all green.

PR #239, opened 2026-09-02
Makefile: link vendored hidapi into the Windows UI archive

pedromessetti:
Fixes `make fyne-ui-windows` failing with undefined references to `hid_*`.

## Problem
The Windows UI cross-build compiles `cm108_ptt.o` with `-DHAVE_HIDAPI` (via the `OS=Windows_NT` `internal_deps` sub-make), but `libmercury_core_w64.a` omitted the vendored `radio_io/hidapi-w64/hid.o`. The link step then failed:

```
undefined reference to `hid_write`
undefined reference to `hid_init`
...
```

## Fix
- Resolve `HIDAPI_W64_OBJ` directly (guarded by the same `hid.c` wildcard check as the detection block), since `HIDAPI_OBJS` is empty in the host (Linux) context.
- Add it to the `libmercury_core_w64.a` prerequisites and archive.
- Build the vendored `hid.o` with the mingw compiler so the cross-build produces a COFF object instead of a native ELF one.

Verified: `make fyne-ui-windows` now produces `windows-installer/mercury-ui.exe`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -229,8 +229,12 @@ main.o: main.c .git_hash_stamp
 	$(CC) $(CFLAGS) -c main.c
 
 # Vendored hidapi for the Windows cross-build (see the detection block above).
+# Built with $(MINGW_CC) rather than $(CC) so it is correct both for the native
+# `windows` target (where CC is already $(MINGW_CC)) and for the
+# `fyne-ui-windows` cross-build (whose host context would otherwise compile a
+# native ELF object that the mingw linker rejects).
 $(HIDAPI_W64_DIR)/hid.o: $(HIDAPI_W64_DIR)/src/hid.c
-	$(CC) -O2 -I$(HIDAPI_W64_DIR)/include -I$(HIDAPI_W64_DIR)/src -c $< -o $@
+	$(MINGW_CC) -O2 -I$(HIDAPI_W64_DIR)/include -I$(HIDAPI_W64_DIR)/src -c $< -o $@
 
 # Vendored hidapi for macOS: the IOKit backend, built from source so a stock
 # macOS with only the Xcode command line tools still gets CM108 PTT.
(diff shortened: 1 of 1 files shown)


rafael2k:
Correct on both counts, and I reproduced the failure before and after.

**`hid.o` built with `$(CC)`** — right diagnosis. On the `windows` target `CC` is already the cross compiler so it happened to work; from `fyne-ui-windows` the host context compiles an ELF object and the mingw linker rejects it. Confirmed the object is now a PE/COFF file:

```
radio_io/hidapi-w64/hid.o: x86-64 COFF object file
```

**`hid.o` missing from the archive** — also real, and it bites regardless of how it was built. `internal_deps OS=Windows_NT` compiles `cm108_ptt.o` with `-DHAVE_HIDAPI`, so the archive carries unresolved `hid_*`. On my branch, before your fix:

```
$ x86_64-w64-mingw32-nm libmercury_core_w64.a | grep '^ *U hid_'
                 U hid_close
                 U hid_enumerate
                 U hid_exit
                 U hid_free_enumeration
                 U hid_init
```

and with your branch checked out they resolve. The reason `$(HIDAPI_OBJS)` can't serve — it is evaluated in the host context where it is empty on Linux — is worth the comment you gave it; that is exactly the kind of thing that gets "simplified" later.

**Merge-ready.** One coordination note: this collides with #238, which edits the same `libmercury_core_w64.a` rule to add the RaptorQ objects (the bridge calls `bcast_file_*`). Whichever lands second needs a trivial rebase; the merged rule wants both:

```make
$(MINGW_AR) rcs $@ $(MERCURY_CORE_OBJS_W64) $(HIDAPI_W64_OBJ) $(BCAST_FILE_OBJS_W64) \
    $(FYNE_UI_DIR)/engine/mercury_bridge_w64.o
```

Happy to take the rebase on #238 if you merge this first — it is the smaller change and unblocks the Windows UI on its own.

PR #240, opened 2026-09-02
ui: allow runtime ALSA/PulseAudio subsystem switch on Linux

pedromessetti:
Add an Audio Subsystem selector to the Fyne Soundcards dialog that is
only shown when the engine reports a changeable subsystem (Linux). The
selection is sent as value4 of set_audio_config; the engine restarts
just the audioio subsystem (not the whole engine) and persists the new
sound_system to mercury.ini. Device ids are dropped on a switch so the
new backend resolves its own default device.

### DIFF
--- a/audioio/audioio.c
+++ b/audioio/audioio.c
@@ -2387,13 +2387,13 @@ int audioio_restart(const char *capture_dev, const char *playback_dev,
     else
         capture_input_channel_layout = LEFT;
 
-    if (capture_dev && capture_dev[0] != '\0')
+    if (capture_dev)
     {
         strncpy(s_capture_dev, capture_dev, sizeof(s_capture_dev) - 1);
         s_capture_dev[sizeof(s_capture_dev) - 1] = '\0';
     }
 
-    if (playback_dev && playback_dev[0] != '\0')
+    if (playback_dev)
     {
         strncpy(s_playback_dev, playback_dev, sizeof(s_playback_dev) - 1);
         s_playback_dev[sizeof(s_playback_dev) - 1] = '\0';
--- a/audioio/audioio.c
+++ b/audioio/audioio.c
@@ -228,6 +228,36 @@ static inline uint64_t audioio_monotonic_ms(void)
 #endif
 }
 
+void audioio_health_reset(void)
+{
+    pthread_mutex_lock(&s_health_lock);
+    s_cap_health  = AUDIO_HEALTH_STOPPED;
+    s_play_health = AUDIO_HEALTH_STOPPED;
+    s_health_reason[0] = '\0';
+    pthread_mutex_unlock(&s_health_lock);
+}
+
+int audioio_wait_healthy(int timeout_ms)
+{
+    if (timeout_ms < 0)
+        timeout_ms = 0;
+
+    uint64_t deadline = audioio_monotonic_ms() + (uint64_t)timeout_ms;
+    for (;;)
+    {
+        audio_health_t c = audioio_capture_health();
+        audio_health_t p = audioio_playback_health();
+
+        if (c == AUDIO_HEALTH_FAILED || p == AUDIO_HEALTH_FAILED)
+            return -1;
+        if (c == AUDIO_HEALTH_RUNNING && p == AUDIO_HEALTH_RUNNING)
+            return 0;
+        if (audioio_monotonic_ms() >= deadline)
+            return -2;
+        ffthread_sleep(5);
+    }
+}
+
 #if defined(__linux__)
 static pthread_mutex_t s_pulse_lock = PTHREAD_MUTEX_INITIALIZER;
 
@@ -992,7 +1022,6 @@ void *radio_playback_thread(void *device_ptr)
     format_device_display(audio_subsystem, FFAUDIO_DEV_PLAYBACK,
                            device_ptr ? (const char *)device_ptr : NULL,
                            play_dev_display, sizeof(play_dev_display));
-    audio_health_set(false, AUDIO_HEALTH_RUNNING, NULL);
     HLOGI("audio-play", "I/O playback (%s) %s / %dHz / %dch / %dms buffer",
           play_dev_display,
           cfg->format == FFAUDIO_F_FLOAT32 ? "float32" :
@@ -1017,6 +1046,8 @@ void *radio_playback_thread(void *device_ptr)
         HLOGE("audio-play",
               "Device negotiated %d channels; only mono/stereo playback is supported, aborting",
               cfg->channels);
+        audio_health_set(false, AUDIO_HEALTH_FAILED,
+                         "device negotiated more than 2 playback channels");
         goto cleanup_play;
     }
 
@@ -1110,6 +1141,10 @@ void *radio_playback_thread(void *device_ptr)
         resampler_init_up(resample_ratio);
         resamp_up_reset(&up_rs);
     }
+
+    /* Reached only after the device, rate, format and channel checks above:
+     * the playback path is fully validated and about to stream. */
+    audio_health_set(false, AUDIO_HEALTH_RUNNING, NULL);
 
     while (!shutdown_ && !audio_shutdown_)
     {
@@ -1393,7 +1428,6 @@ void *radio_capture_thread(void *device_ptr)
     format_device_display(audio_subsystem, FFAUDIO_DEV_CAPTURE,
                            device_ptr ? (const char *)device_ptr : NULL,
                            cap_dev_display, sizeof(cap_dev_display));
-    audio_health_set(true, AUDIO_HEALTH_RUNNING, NULL);
     HLOGI("audio-cap", "I/O capture (%s) %s / %dHz / %dch / %dms buffer",
           cap_dev_display,
           cfg->format == FFAUDIO_F_FLOAT32 ? "float32" :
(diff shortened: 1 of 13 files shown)


rafael2k:
Read through; the design holds together and the one semantic change is safe. Notes rather than objections.

**Blanking the device ids on a subsystem switch is the right call.** Device identifiers are subsystem-specific, so carrying an ALSA `hw:1,0` into PulseAudio would either fail or, worse, resolve to something unintended. Dropping them so the new subsystem picks its own default is correct, and the comment says why.

**The `audioio_restart` change deserves a callout in the commit message.** This:

```c
-    if (capture_dev && capture_dev[0] != '\0')
+    if (capture_dev)
```

changes what an empty string *means* — from "keep the current device" to "clear it". That is required for the switch to work, and it is safe here because `ui_communication.c:191` is the only caller and it is the one you just taught to pass empty deliberately. But it is a contract change on a public-ish function, and the next caller added will not know. Worth a line in `audioio.h` saying empty clears and NULL keeps.

**Republishing the device list on switch** (`soundcard_list_pending = 1`) is a good catch — the remote UI would otherwise show the previous subsystem's devices as if they were selectable.

Two questions, neither blocking:

1. What happens if the new subsystem fails to open? The operator has switched away from a working configuration; does `audioio_restart` leave the old one running, or is the station deaf until they switch back? If it is the latter, a fallback to the previous subsystem on failure would be worth having — that failure mode is unattended-station-shaped.

2. Is `ctx->audio_system` read anywhere off the UI thread while the restart runs? I did not spot a race, but the audio threads are being torn down concurrently and it is the kind of thing that only shows up on a slow Pi.

`cfg_sound_system_parse()` rejecting an unknown name rather than defaulting is the right choice — a typo in a config should be visible, not silently ALSA.

Merge-ready from my side once you have clicked through a switch on a real box; the logic is right but this is the sort of change where the interesting failures are in the device layer rather than the control flow.

pedromessetti:
Thanks for the thorough review — all fair points, and the two concrete ones are now pushed.

**`audioio_restart` contract** — agreed. Documented on `audioio_restart` in `audioio.h`: `NULL` = keep current, empty string = clear/use the subsystem default, non-empty = use this device, plus the "failed switch leaves the audio path stopped" caveat.

**Failure fallback (Q1)** — you read it correctly. `audioio_restart` stops and joins the old threads before starting the new subsystem, so if the new backend fails to open, the old configuration is gone and the station is deaf until the operator switches back. The engine survives and `audioio_health_*` surfaces the error, but there's no automatic rollback. A "revert to last-good config on failure" is worth having, but it needs a success/failure signal plumbed back from the restart to the command handler — I'd rather track that as a follow-up than grow this PR.

**Thread-safety (Q2)** — you were right, it's a genuine race: the command handler writes `ctx->audio_system` while `ui_publisher_thread` reads it. Fixed by making it `_Atomic int` and load/store-ing explicitly at every access.

**`cfg_sound_system_parse` rejecting unknowns** — agreed; a config typo should be loud, not silently ALSA.

rafael2k:
Both follow-ups look right — the `_Atomic` conversion covers every access I can find, and the `audioio_restart` contract comment is exactly the missing piece.

On Q1 I want to narrow the follow-up, because there is a smaller part of it that I do not think should wait. As it stands:

```c
audioio_restart(ctx->selected_capture_dev, ctx->selected_playback_dev,
                active_system, ctx->rx_input_channel);
HLOGI(UI_LOG_TAG, "Audioio subsystem restarted successfully");

// Persist audio config to INI
...
ctx->cfg.sound_system = active_system;
if (ctx->cfg_path[0] && cfg_write(&ctx->cfg, ctx->cfg_path))
```

So a switch to a subsystem that cannot open **writes itself into mercury.ini anyway**. The operator is then not one click from recovery, as your reply describes — they restart and come up deaf again, because the saved configuration is now the broken one. That turns a transient mistake into a persistent one, and on a remote station the UI is exactly what they have just lost the ability to hear through.

The unconditional `"restarted successfully"` compounds it: the log says the thing worked at the moment it did not, which is the log line someone will be reading while diagnosing.

The full rollback needs the plumbing you describe and is fair to defer. But **not persisting a configuration that demonstrably failed** does not — `audioio_health_ok()` is already there and is how the header says failure is reported:

```c
audioio_restart(...);

char why[192];
if (!audioio_health_ok(why, sizeof(why))) {
    HLOGE(UI_LOG_TAG, "Audio subsystem %s failed to start: %s",
          cfg_sound_system_name(active_system), why);
    /* leave mercury.ini alone: a config that does not work must not
     * become the one we come back up with */
} else {
    HLOGI(UI_LOG_TAG, "Audioio subsystem restarted successfully");
    ...persist...
}
```

(with whatever settle the health flag needs to be meaningful — you know that path better than I do).

Everything else is merge-ready as far as I am concerned. This one is small, and it is the difference between "switch back" and "reflash the config by hand".

pedromessetti:
Agreed — that smaller part shouldn't wait, and it's now fixed.

`set_audio_config` no longer logs "restarted successfully" or writes `mercury.ini` unconditionally. `audioio_restart` now resets the health flags before spawning the new threads, and a new `audioio_wait_healthy(timeout_ms)` lets the handler wait for the capture/playback paths to reach RUNNING or FAILED (bounded, so a slow PulseAudio autospawn can't wedge the command handler). Persist happens only on a healthy start; on FAILED we log the health reason and leave `mercury.ini` alone.

One wrinkle a naive `audioio_health_ok()` check would have missed: health was being set to RUNNING immediately after `open()`, before the rate/format/channel validation that can still fail (and the channel check didn't set FAILED at all). I moved the RUNNING set to after all of that, so RUNNING means the path is actually streaming — otherwise a rate-negotiation failure would have been reported as healthy and persisted. The full rollback-to-last-good still needs the restart-to-handler success/failure plumbing, which I'm keeping as the deferred follow-up.

rafael2k:
`5f363b8` does the job, and the extra piece you added is better than what I suggested.

I proposed calling `audioio_health_ok()` after the restart. That would have been **wrong**, and your version catches why: without `audioio_health_reset()` the flags still describe the *old* subsystem, so a healthy previous configuration would have masked the failure of the new one. Resetting before the threads spawn, then waiting for RUNNING-or-FAILED, is the correct shape.

Moving the RUNNING flag past the device/rate/format/channel checks is the part I would not have thought to ask for:

```c
/* Reached only after the device, rate, format and channel checks above:
 * the playback path is fully validated and about to stream. */
audio_health_set(false, AUDIO_HEALTH_RUNNING, NULL);
```

RUNNING now means streaming rather than merely opened — which matters well beyond this PR, since it is also what the UI's audio-health indicator reports. A card that opens and then rejects the rate was previously "healthy".

Built and ran the suite on `5f363b8`: clean build, all unit tests pass.

Two small things, neither blocking:

- `audioio_wait_healthy(3000)` — 3 s is generous for a local device and short for a PulseAudio daemon that is being started on demand. If a slow-but-successful start ever trips it, the operator gets "failed" *and* a working audio path, which is a confusing pair. Worth a log line distinguishing the timeout return (`-2`) from a real failure (`-1`) if it is not already there.
- The unattended case is now safe in the way that matters — a bad switch no longer survives a restart. Automatic rollback is still worth having eventually, but it is genuinely follow-up now rather than a gap.

**Merge-ready from my side.**

One design note for later, not for this PR: the switch is Linux-only, but the platform matrix is Linux 3 (ALSA / PulseAudio / OSS), Windows 2 (WASAPI / DirectSound), macOS 1 (CoreAudio), Android 1 (AAudio, in future). Windows genuinely has a choice to offer — DirectSound is the fallback for drivers WASAPI does not get on with. Driving the picker off "how many subsystems did this build compile in" rather than `#ifdef __linux__` would cover Windows for free and hide it on macOS, where offering it is just noise.

pedromessetti:
Thanks — and thanks for actually building and running it.

On the timeout: the distinction is already there. `-1` logs

```
HLOGE("Audio subsystem %s failed to start: %s", ..., <health reason>)
```

while `-2` logs

```
HLOGW("Audio subsystem %s did not report health in time; not persisting config")
```

So a slow-but-successful start reads as "indeterminate, nothing written to disk" — not "failed" — and the (working) path can be re-applied to persist it. The 3 s is the settle ceiling rather than the typical wait: ALSA and an already-running PulseAudio settle well under it, and the bound just caps how long the command handler blocks.

Rollback: agreed, tracked as follow-up.

On the picker design note: good call. Driving it off "how many switchable subsystems this build compiled in" instead of `#ifdef __linux__` would get Windows (WASAPI/DirectSound) the picker for free and hide it on macOS/Android. I'll fold that into a follow-up rather than this PR.

rafael2k:
You are right, and I should have checked before raising it — both branches are there:

```
HLOGE("Audio subsystem %s failed to start: %s", ...)          /* -1 */
HLOGW("Audio subsystem %s did not report health in time; ...") /* -2 */
```

A slow-but-successful start reads as *indeterminate, nothing written* rather than *failed*, which is the right call and better than what I suggested. My concern was already handled; I asked about it without looking.

Rollback and the picker-by-compiled-subsystems both make sense as follow-ups.

**Merge-ready from my side.** For the record, what I verified rather than read: built `5f363b8` clean and ran the full unit suite (green), and confirmed the `_Atomic` conversion covers every `ctx->audio_system` access.

One note now that #238 has landed on trunk: it touches `gui_interface/ui_communication.c` too (the broadcast file panel's status plumbing), so you may get a small conflict on rebase. Nothing structural — different functions in the same file.

PR #242, opened 2026-09-02
Makefile: stop archiving rigctl_parse.o twice for Windows

rafael2k:
Follow-up to #239 — small, and it is the only thing I found still wrong after that landed.

`MERCURY_CORE_OBJS_W64` starts as a copy of `MERCURY_CORE_OBJS`, which already contains `radio_io/rigctl_parse.o` when the **host** has hamlib, and then adds it again for the cross build. The Windows archive carried two identical members:

```
$ x86_64-w64-mingw32-ar t libmercury_core_w64.a | sort | uniq -d
rigctl_parse.o
```

GNU ld tolerates it, which is why nothing has failed — but it is wasted space and a duplicate-symbol hazard with a stricter linker.

Adds it only when not already present. Verified both host configurations, since the point is not to break the case the unconditional `+=` exists for:

| | trunk | fixed |
|---|---|---|
| host **with** hamlib | 2 | 1 |
| host **without** hamlib | 1 | 1 |

`MERCURY_CORE_OBJS_W64` also becomes simply-expanded (`:=`), because the dedupe refers to the variable being assigned — with the recursive form make refuses outright (*"Recursive variable references itself"*). That is worth knowing if anyone extends this block later.

Checked from clean: native build, unit suite, Go tests, `make windows`, and the Windows archive — one `rigctl_parse.o`, no duplicated members, `hid.o` still present, and every `hid_*` referenced still resolved within the archive.

While I was there I also confirmed #239 does what it set out to: `make fyne-ui-windows` now completes and produces `mercury-ui.exe` (27 MB).


rafael2k:
Rebased onto trunk now that #238 has landed (both touch the `libmercury_core_w64.a` rule; the merge kept the dedupe alongside its `BCAST_FILE_OBJS_W64`).

Re-verified everything on the merged tree rather than relying on the pre-rebase run:

```
host HAVE_HAMLIB=1 -> rigctl_parse.o in ar line: 1
host HAVE_HAMLIB=0 -> rigctl_parse.o in ar line: 1

archive:  0 duplicated members
          1 rigctl_parse.o,  1 hid.o,  5 raptorq objects
          every hid_* referenced is also defined within the archive

native from clean       rc=0, full unit suite green
make windows from clean rc=0
make fyne-ui-windows    rc=0 -> mercury-ui.exe, 27357696 bytes
```

That last line is the one worth having: it confirms the Windows UI still links end to end after both changes, which is what #239 set out to fix in the first place.

PR #245, opened 2026-09-03
ci: build Fyne UI (native + Windows) on every PR and mercuryv2 push

pedromessetti:
Adds two jobs to the Build workflow so every PR and push to `mercuryv2` verifies the Fyne UI still builds:

- `build-fyne-ui-linux` — `make fyne-ui` (native, Debian 13) with Fyne's OpenGL/X11/Wayland dev headers
- `build-fyne-ui-windows` — `make fyne-ui-windows` (MinGW cross-compile)

Both use a plain checkout (no ssh-key) so PRs from forks are covered, and are gated by new `workflow_dispatch` inputs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -339,9 +339,12 @@ jobs:
   # engine via libmercury_core.a.  Fyne's GLFW frontend is vendored and builds
   # BOTH X11 and Wayland backends by default, so the container needs their dev
   # headers (xorg-dev, libwayland-dev, libxkbcommon-dev) alongside the C core
-  # deps (libpulse/libasound/libhamlib).  Plain checkout (no ssh-key) so PRs
-  # from forks -- exactly the contributor whose GUI break we most want caught --
-  # are covered; secrets are not available to those runs.
+  # deps (libpulse/libasound/libhamlib).  libhidapi-dev is what turns this from
+  # "the UI compiles" into "the UI links the way it ships": with it, the
+  # Makefile sets HAVE_HIDAPI and cm108_ptt.o references hid_*, so the link
+  # really exercises the CGO_LDFLAGS="$(HIDAPI_LDFLAGS)" path.  Plain checkout
+  # (no ssh-key) so PRs from forks -- exactly the contributor whose GUI break we
+  # most want caught -- are covered; secrets are not available to those runs.
   build-fyne-ui-linux:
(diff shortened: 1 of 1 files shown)


rafael2k:
Yes, we need this — and it is the gap that let the last two GUI breakages reach trunk. Neither `make fyne-ui` nor `make fyne-ui-windows` was built anywhere in CI, so #239's Windows link failure and a transport bug of mine both got through green matrices. Structure looks right: plain checkout so fork PRs are covered, `workflow_dispatch` inputs consistent with the existing jobs, and a real artefact check rather than trusting the exit code.

Two things I would change, one of which is a blind spot rather than a nit.

**1. Add `libhidapi-dev` to the native job.** Without it `pkg-config` finds no hidapi, so the Makefile leaves `HAVE_HIDAPI` unset, `cm108_ptt.c` compiles its non-hidapi path, and there are no `hid_*` references to resolve at all:

```
$ gcc -c radio_io/cm108_ptt.c ... && nm cm108_ptt.o | grep -c ' U hid_'
0
```

So the job links whether or not `fyne-ui` passes `CGO_LDFLAGS="$(HIDAPI_LDFLAGS)"` — meaning a regression that removed it would still go green. That line exists precisely because the build fails without it on any host that *has* hidapi, which is every developer machine and the release builder. One package turns this job from "the UI compiles" into "the UI links the way it ships".

**2. Go toolchain.** `gui_interface/fyne-ui/go.mod` requires `go 1.26.8`; trixie ships `golang-go 2:1.24~2`. It works — `GOTOOLCHAIN=auto` fetches 1.26.8 from the proxy — but it means every PR downloads a toolchain, and a proxy hiccup surfaces as a confusing build error rather than "cannot get Go 1.26.8". Worth either setting `GOTOOLCHAIN` explicitly so the intent is visible, or installing upstream Go in the container. (That requirement is mine — I bumped the module — so the cost is my doing, not this PR's.)

Neither blocks it. Even as-is this catches the class of failure we have actually been hitting.

For the record, I have run both targets locally on this branch's parent: `make fyne-ui` produces `mercury-ui`, and `make fyne-ui-windows` produces `windows-installer/mercury-ui.exe` (27 MB), so the targets themselves are sound in a clean tree.

On #240 — merge it. Reviewed and confirmed: the `_Atomic` conversion covers every access, the health-gate stops a failed subsystem being written to `mercury.ini`, and the `-1`/`-2` logging distinction was already there when I asked about it.

PR #246, opened 2026-09-03
ui: derive the audio subsystem picker from the compiled-in backends

pedromessetti:
Follow up for PR #240 

Replace the #ifdef __linux__ subsystem list with audioio_available_subsystems(),
which reports the backends this build actually compiled in, in display order:
Linux (ALSA/PulseAudio/OSS when built), Windows (WASAPI/DirectSound), macOS
(CoreAudio), Android (AAudio). A single-entry list keeps the picker hidden on
macOS/Android while Windows now gets the WASAPI/DirectSound choice for free.

The websocket audio_system message and the embedded UI's AudioSubsystems()
both read this list, so the picker is driven by the build rather than by the
UI's own platform.

### DIFF
--- a/audioio/audioio.c
+++ b/audioio/audioio.c
@@ -312,6 +312,35 @@ int audioio_pick_default_subsystem(void)
 #endif
 }
 
+int audioio_available_subsystems(int *subsystems, int max)
+{
+    int n = 0;
+    if (!subsystems || max <= 0)
+        return 0;
+
+#define ADD(s) do { if (n < max) subsystems[n] = (s); n++; } while (0)
+#if defined(__linux__)
+    ADD(AUDIO_SUBSYSTEM_ALSA);
+    ADD(AUDIO_SUBSYSTEM_PULSE);
+    if (OSS_IFACE != NULL)
+        ADD(AUDIO_SUBSYSTEM_OSS);
+#elif defined(_WIN32)
+    ADD(AUDIO_SUBSYSTEM_WASAPI);
+    ADD(AUDIO_SUBSYSTEM_DSOUND);
+#elif defined(__FreeBSD__)
+    ADD(AUDIO_SUBSYSTEM_OSS);
+#elif defined(__APPLE__)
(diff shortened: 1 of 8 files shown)


rafael2k:
This is the shape I had in mind, and better in one respect I had not thought about: gating OSS on `OSS_IFACE != NULL` means the list reflects what the build *actually linked*, not just the target platform — so a Linux build without OSS4 headers correctly offers two entries rather than advertising a backend that is not there.

Windows getting WASAPI/DirectSound for free is the point of doing it this way, and a single-entry list hiding the picker on macOS and Android is the right default: offering a choice of one is just noise.

**One latent trap worth closing while it is cheap.** `ADD` counts entries it did not write:

```c
#define ADD(s) do { if (n < max) subsystems[n] = (s); n++; } while (0)
```

so the return value is "how many exist", which can exceed `max`. The caller then uses it as an iteration bound over the array it passed:

```c
int subsystems[8];
int n = audioio_available_subsystems(subsystems, 8);
...
for (int i = 0; i < n; i++)          /* reads subsystems[i] */
```

Today that cannot overflow — the largest list is three (Linux with OSS) — so this is not a live bug. But it is the snprintf contract (return what was *needed*) wired to a loop that assumes the return is what was *written*, and it only bites when someone adds a ninth backend, at which point it is a silent stack over-read rather than a compile error.

Either return the written count:

```c
return n < max ? n : max;
```

or clamp at the call site. Nothing currently wants "how many would fit", so returning the written count is the simpler of the two and removes the question.

Everything else reads correctly to me — the websocket path and `AudioSubsystems()` both taking the same list is what stops the UI and the engine disagreeing about which platform they are on.

Worth noting for whoever merges: #244 landed on trunk and touches `gui_interface/fyne-ui/mercury_chat_window.go` and `mercury-client/client/client.go`, neither of which this PR touches, so no conflict expected.

pedromessetti:
Done — `ADD` now only advances `n` when it actually writes a slot, so the return value is the written count and callers can safely treat it as the array bound. Pushed as `3d42c43`.

PR #253, opened 2026-09-04
Persist ARQ/broadcast chat history across restarts

pedromessetti:
## Summary

Adds a persistent message store so the ARQ and broadcast chat messages a station exchanges are **not lost when the application is closed and reopened**. Today the Fyne chat window keeps only a 200-message in-memory ring, so history vanishes on exit.

## What it does

- **New engine module** `common/message_store.{c,h}`: appends each exchanged chat message as a JSONL record to a file and keeps a bounded in-memory ring. It reloads the file on startup, so history survives a restart.
- **Message identification** is hooked at the four engine chokepoints:
  - ARQ RX (`arq.c:cb_deliver_rx_data`) and ARQ TX (`arq.c:arq_submit_tcp_payload`)
  - Broadcast TX (`tcp_interfaces.c:bcast_process_decoded_frame`) and RX (`tcp_interfaces.c:send_thread`)
- Only **printable text lines** (≤1024 chars) are stored, so file transfers and AX.25/Reticulum binary frames are filtered out of the history.
- **New `HISTORY` TNC command** (documented in `docs/TNC.md`) returns the persisted history as `HISTORYMSG <json>` lines between `HISTORY <n>` / `HISTORYEND`.
- **Fyne UI + mercury-client**: `GetHistory()`/`History()` fetch the history on connect and repopulate both chat panes.

## Configuration

New `[store]` section in `mercury.ini` (`mercury.ini.example`):

```ini
[store]
enabled = true          ; default
path =                  ; empty = platform data dir (~/.local/state/mercury/messages.jsonl on Linux)
max_messages = 500
```

## JSONL record schema

```json
{"ts":1699999999,"ms":123,"plane":"arq","dir":"rx","peer":"CALL","text":"hello"}
```

## Tests

- New `tests/common/test_message_store.c` (7 tests: append/get round-trip, newline splitting, binary/long-line filtering, persistence reload, ring capacity, out-of-range).
- `HISTORY` command framing tests added to `test_tcp_interfaces.c`.

## Notes / known limitations

- **"Chat text only" is a heuristic on the ARQ data port**: it shares a pipe with file transfers, so a *text* file transfer is indistinguishable from chat. Binary transfers are cleanly filtered (control bytes + length cap).
- **Broadcast storage happens while a broadcast client is connected** (hooked at the delivery/framing points). Let me know if headless broadcast capture is wanted too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,7 @@ MERCURY_LINK_INPUTS = \
 	datalink_arq/arq_fsm.o datalink_arq/arq_protocol.o datalink_arq/arq_timing.o datalink_arq/arq_modem.o \
 	datalink_broadcast/broadcast.o datalink_broadcast/kiss.o modem/modem.o modem/framer.o modem/channel_busy.o modem/freedv/libfreedvdata.a \
 	audioio/audioio.a common/os_interop.o common/ring_buffer_posix.o common/shm_posix.o common/crc6.o common/hermes_log.o common/virtual_clock.o \
-	common/chan.o common/queue.o common/mercury_engine.o common/mercury_cli.o data_interfaces/tcp_interfaces.o data_interfaces/net.o \
+	common/chan.o common/queue.o common/mercury_engine.o common/mercury_cli.o common/message_store.o data_interfaces/tcp_interfaces.o data_interfaces/net.o \
 	gui_interface/ui_communication.o gui_interface/ui_status.o gui_interface/ui_devices.o \
 	gui_interface/websocket/mongoose.o gui_interface/websocket/mercury_websocket.o \
 	gui_interface/websocket/web_packed.o \
@@ -297,7 +297,7 @@ MERCURY_CORE_OBJS = \
 	datalink_broadcast/broadcast.o datalink_broadcast/kiss.o \
 	modem/modem.o modem/framer.o modem/channel_busy.o \
 	common/os_interop.o common/ring_buffer_posix.o common/shm_posix.o common/crc6.o common/hermes_log.o common/virtual_clock.o \
-	common/chan.o common/queue.o common/mercury_engine.o common/mercury_cli.o \
+	common/chan.o common/queue.o common/mercury_engine.o common/mercury_cli.o common/message_store.o \
 	data_interfaces/tcp_interfaces.o data_interfaces/net.o \
 	gui_interface/ui_communication.o gui_interface/ui_status.o gui_interface/ui_devices.o \
 	gui_interface/websocket/mongoose.o gui_interface/websocket/mercury_websocket.o \
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,7 @@ MERCURY_LINK_INPUTS = \
 	datalink_broadcast/broadcast.o datalink_broadcast/kiss.o modem/modem.o modem/framer.o modem/channel_busy.o modem/freedv/libfreedvdata.a \
 	audioio/audioio.a common/os_interop.o common/ring_buffer_posix.o common/shm_posix.o common/crc6.o common/hermes_log.o common/virtual_clock.o \
 	common/chan.o common/queue.o common/mercury_engine.o common/mercury_cli.o common/message_store.o data_interfaces/tcp_interfaces.o data_interfaces/net.o \
-	gui_interface/ui_communication.o gui_interface/ui_status.o gui_interface/ui_devices.o \
+	gui_interface/ui_communication.o gui_interface/ui_status.o gui_interface/ui_devices.o gui_interface/ui_history.o \
 	gui_interface/websocket/mongoose.o gui_interface/websocket/mercury_websocket.o \
 	gui_interface/websocket/web_packed.o \
 	radio_io/radio_io.o radio_io/serial_ptt.o radio_io/cm108_ptt.o $(HIDAPI_OBJS)
@@ -299,7 +299,7 @@ MERCURY_CORE_OBJS = \
 	common/os_interop.o common/ring_buffer_posix.o common/shm_posix.o common/crc6.o common/hermes_log.o common/virtual_clock.o \
 	common/chan.o common/queue.o common/mercury_engine.o common/mercury_cli.o common/message_store.o \
 	data_interfaces/tcp_interfaces.o data_interfaces/net.o \
-	gui_interface/ui_communication.o gui_interface/ui_status.o gui_interface/ui_devices.o \
+	gui_interface/ui_communication.o gui_interface/ui_status.o gui_interface/ui_devices.o gui_interface/ui_history.o \
 	gui_interface/websocket/mongoose.o gui_interface/websocket/mercury_websocket.o \
 	gui_interface/websocket/web_packed.o \
 	radio_io/radio_io.o radio_io/serial_ptt.o radio_io/cm108_ptt.o $(HIDAPI_OBJS)
--- a/common/Makefile
+++ b/common/Makefile
@@ -16,7 +16,7 @@ ENGINE_CFLAGS = $(CFG_CFLAGS) -I.. -I../modem -I../modem/freedv -I../datalink_br
 
 all: os_interop.o ring_buffer_posix.o shm_posix.o crc6.o hermes_log.o chan.o queue.o \
      virtual_clock.o cfg_utils.o iniparser/iniparser.o iniparser/dictionary.o watterson.o mercury_engine.o \
-     mercury_cli.o
+     mercury_cli.o message_store.o
 
 os_interop.o: os_interop.c
 	$(CC) $(CFLAGS) -c os_interop.c
@@ -59,6 +59,9 @@ mercury_engine.o: mercury_engine.c mercury_engine.h
 
 mercury_cli.o: mercury_cli.c mercury_cli.h
 	$(CC) $(ENGINE_CFLAGS) $(CLI_SHM_CFLAGS) -c mercury_cli.c
+
+message_store.o: message_store.c message_store.h
+	$(CC) $(CFLAGS) -c message_store.c
 
 .PHONY: all clean
 
--- a/common/cfg_utils.c
+++ b/common/cfg_utils.c
@@ -30,6 +30,7 @@
 #include "../data_interfaces/tcp_interfaces.h"
 #include "../gui_interface/ui_communication.h"
 #include "../datalink_arq/arq_protocol.h"
+#include "message_store.h"
 
 void cfg_set_defaults(mercury_config *cfg)
 {
@@ -76,6 +77,9 @@ void cfg_set_defaults(mercury_config *cfg)
     cfg->busy_hysteresis_db          = 3;
     cfg->busy_on_debounce_ms         = 300;
     cfg->busy_hang_ms                = 1500;
+    cfg->store_enabled               = true;
+    cfg->store_path[0]               = '\0';
+    cfg->store_max_messages          = MSG_STORE_DEFAULT_MAX_LINES;
 }
 
 /* Map a sound-system name to the AUDIO_SUBSYSTEM_* constant.
@@ -410,6 +414,18 @@ bool cfg_read(mercury_config *cfg, const char *ini_path)
     i = iniparser_getint(ini, CFG_KEY_TX_DELAY_MS, cfg->tx_delay_ms);
     if (i >= 0 && i <= 2000) cfg->tx_delay_ms = i;
 
+    cfg->store_enabled = (bool) iniparser_getboolean(ini, CFG_KEY_STORE_ENABLED,
+                                                     cfg->store_enabled ? 1 : 0);
+
+    s = iniparser_getstring(ini, CFG_KEY_STORE_PATH, NULL);
+    if (s) {
+        strncpy(cfg->store_path, s, sizeof(cfg->store_path) - 1);
+        cfg->store_path[sizeof(cfg->store_path) - 1] = '\0';
+    }
+
+    i = iniparser_getint(ini, CFG_KEY_STORE_MAX_MESSAGES, cfg->store_max_messages);
+    if (i >= 1 && i <= 10000) cfg->store_max_messages = i;
+
     iniparser_freedict(ini);
     return true;
 }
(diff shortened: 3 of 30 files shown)


rafael2k:
Reviewed `store-messages` @ 23cbee5. Built it and ran the new test (7/7 pass), and reproduced one of these with a standalone probe. Nice module overall — the JSON escaping is correct and conservatively bounded, the store lock is narrow, over-long lines are discarded whole rather than truncated into broken records, and the binary filter keeps file transfers and AX.25/Reticulum frames out. Two things below look like they need fixing before this lands.

---

## HIGH — `HISTORY` bulk-dumps through the lossy control writer

`data_interfaces/tcp_interfaces.c:476`

The CTL client socket is non-blocking (`tcp_interfaces.c:819`), and `tcp_write()` (`data_interfaces/net.c:235`) treats a short send as fatal — `net_set_status(CTL_TCP_PORT, NET_RESTART)` — and silently drops the line on `EAGAIN`. Its own comment says control writes are "periodic status messages where dropping one beats blocking the reactor". `tcp_write_all()` exists for exactly this case.

`HISTORY` pushes `max_messages` lines in a tight loop from the reactor thread with nothing draining: ~60 KB at the default 500, ~1.2 MB at the documented max of 10000.

So when the send buffer fills mid-dump, either the UI's control connection is reset immediately after connect, or `HISTORYEND` is dropped, `GetHistory` blocks its full 10 s and returns "timed out", and the whole history is lost.

Suggest `tcp_write_all()` for the dump.

## HIGH — the partial-line feed buffer is never reset between sessions

`common/message_store.c:311`

There is no reset entry point in `message_store.h`, and nothing is called on ARQ connect/disconnect or broadcast client connect. Residue from one session is glued onto the next message:

```c
msg_store_feed("arq","rx","AAA","hello",5);   /* session ends, no newline */
msg_store_feed("arq","rx","BBB","hi\n",3);
→ {"peer":"BBB","text":"hellohi"}             /* one record, wrong peer, wrong text */
```

It bites harder on the broadcast RX hook (`tcp_interfaces.c:1295`): for legacy / hermes-broadcast frames `bcast_get_tx_payload()` returns the **whole padded frame** (`*payload_len_out = (int)frame_size`), so every frame leaves ~100 trailing NUL bytes with no newline in the feed buffer. The next broadcast message is appended to that NUL run, fails `msg_store_is_text()` (NUL < 0x20), and is dropped — meaning **only the first legacy broadcast message per session is ever stored**.

Suggest a `msg_store_reset(plane, dir)` called on session/client boundaries, and feeding only the real payload length on the broadcast path.

---

## MEDIUM

- **`gui_interface/mercury-client/modem/modem.go:324`** — `mc.ARQControlConn.Write` dereferences the field *outside* `mc.mu`. `GetHistory` nil-checks under the lock, then unlocks before using it, unlike `SendCommand` (line 199) which correctly captures `conn := mc.ARQControlConn`. `Disconnect()` (line 418) nils it under the lock — so clicking Disconnect inside the 10 s window is a data race that can panic the Fyne app.
- **`gui_interface/fyne-ui/mercury_chat_window.go:441`** — `loadHistory` runs on every `onConnect`, but `onDisconnect` never clears `arqBox`/`bcastBox`, so a disconnect→connect cycle re-appends the whole history on top of what is already displayed.
- **`gui_interface/mercury-client/modem/modem.go:533`** — `historyRespCh` is `make(chan string, 512)` while `store:max_messages` is documented up to 10000, and the send is `select { case ch <- line: default: }`. Overflow is silently discarded; if the discarded line is `HISTORYEND`, `GetHistory` blocks the full 10 s and errors out. The `HISTORY <n>` header is parsed and thrown away — it could size the channel.
- **`datalink_arq/arq.c:1134`** — the TX hook runs *before* `arq_channel_bus_try_send_payload()`, which returns -1 on `EAGAIN` (chan full) and `EINVAL` (`len > INT_BUFFER_SIZE`) (`arq_channels.c:140`). A message dropped at the bus is still persisted and replayed as if it had gone out. Moving the hook after a successful return fixes it.

## LOW

- **`common/message_store.c`** — no rotation or trimming. `max_messages` bounds only the in-memory ring; the file is append-only forever, and every `msg_store_init` `fgets`-scans the whole file (line 442) just to keep the last `cap` lines. Modest for HF traffic, but the config knob reads like it bounds the file.
- **`common/message_store.c:279`** — one-byte stack overflow in the truncation branch: `n = sizeof(buf) - 1` then `buf[n+1] = '\0'` writes `buf[8192]` on `char buf[8192]`. Unreachable today (`text_len > MSG_MAX_TEXT` returns early, worst case ≈6.3 KB) but live the moment `MSG_MAX_TEXT`/`MSG_MAX_LINE` are retuned. `sizeof(buf) - 2` is the safe clamp.
- **`data_interfaces/tcp_interfaces.c:481`** — TOCTOU across the dump: `msg_store_count()` is sampled once, then each `msg_store_get(i, …)` retakes the lock. A message arriving mid-dump rolls the ring, so the emitted sequence can skip or duplicate, and the real line count then disagrees with the advertised `HISTORY <n>`.
- **`common/message_store.c:442`** — reload has no `fgets` truncation handling; a >8191-byte line is split across calls and each fragment `ring_push`ed as its own "message", so `HISTORY` emits malformed JSON that the Go client silently skips.
- **`datalink_arq/arq.c:307`** — `cb_deliver_rx_data` runs on the ARQ event-loop thread (`arq_fsm.c:537`); the hook adds a `g_conn_lock` acquisition per delivery and `msg_store_emit()` does `fwrite` + `fflush` per line. On the SD-card-backed field stations a flush stall lands on the timing-sensitive turn/deadline loop. Worth deferring the write, or at least dropping the per-line `fflush`.
- **`common/message_store.c:421`** — ring slots are fixed 8192-byte rows: `calloc(max_lines, 8192)` is 4 MB at the default and ~82 MB at the documented max, for records typically under 200 bytes. `calloc` keeps it mostly virtual, so it is a soft cost — but a variable-length store would be materially cheaper on a Pi.

---

The new tests pass, but the `HISTORY` framing tests in `test_tcp_interfaces.c` only cover the stubbed 0- and 1-message cases, so none of the above is caught by the suite. A test that dumps more than a socket buffer's worth would cover the first HIGH, and one that feeds a newline-less fragment then a second message would cover the second.

https://claude.ai/code/session_01JxAhNoNeNS53dA67mCfUg6

rafael2k:
Re-reviewed `store-messages` @ 4f2c1f2. Built it and ran the suite — clean build, all tests pass (the 4 new store tests and the reworked HISTORY framing tests cover the fixes properly, including the large-dump case).

All the earlier findings are fixed, and fixed at the cause rather than worked around:

- `HISTORY` now goes through `tcp_write_all()`, with `tcp_write_all()` generalised to the control port and an early return on a failed write. The `net.c` generalisation keeps the existing DATA semantics intact and adds the `sockfd < 0` guard.
- `msg_store_reset()` is called on ARQ connect *and* disconnect for both directions, and on broadcast client connect; the modem's NUL padding is trimmed before the broadcast RX feed. That closes both halves of the partial-line bug.
- `GetHistory` captures `conn` under the lock like `SendCommand` does, and the extra "already in progress" guard is a good addition.
- The bulk channel is sized from the `HISTORY <n>` header, the TX hook moved after `rc == 0`, the panes are cleared on disconnect, and the emit clamp is `sizeof(buf) - 2`.
- `msg_store_snapshot()` is the right answer to the TOCTOU, and it allocates the summed line lengths rather than `cap * MSG_MAX_LINE`.

Two things I raised earlier were wrong, for the record: the snapshot is not a 4 MB allocation, and a failed `fopen` in the rewrite does not stop persistence — it reopens for append unconditionally.

Two new things, both in the file-bounding rewrite:

## MEDIUM — the rewrite is not crash-safe

`common/message_store.c:257`

`fopen(g_store.path, "w")` truncates the live file and then rewrites the last `cap` lines into it. A crash or power loss during that window loses the whole history, which is the one thing this feature exists to prevent — and the field stations are Pis with no UPS. The local is named `tmp` but points at the real path.

Writing `<path>.tmp` and then `rename()` over the original makes it atomic on POSIX, so a crash leaves either the old file or the new one.

The `fwrite` loop is also unchecked, so a rewrite that fails partway through leaves a short file silently.

## MEDIUM — the rewrite runs on the ARQ event-loop thread

`msg_store_feed()` <- `cb_deliver_rx_data()` <- `arq_fsm.c:761`.

Once per `2 * cap` stored messages that path does `fclose` + `fopen` + `cap` `fwrite`s + `fflush` + `fclose` + `fopen` inline. On the SD-card-backed stations that is a multi-hundred-millisecond stall landing on the thread that carries the ARQ deadlines and turn timers. It is rare when amortised, but it is a spike in a place this codebase has repeatedly been bitten in. Deferring the rewrite (a flag picked up off the event loop, or doing it at init/shutdown only) would keep it off that thread.

## Note, not a defect

`tcp_write_all()` on CTL can now sit in backpressure for up to `WRITE_BACKPRESSURE_TIMEOUT_MS` (5 s). That is the right trade against dropping the dump, but it is new behaviour for the control port: a client that connects and stops reading now delays PTT/DISCONNECTED handling for up to 5 s, where control writes previously never blocked. Bounded, and I would take it over the truncation — just worth it being a deliberate choice.

Neither new finding blocks the feature; both are durability and field-timing rather than correctness of the history itself.

rafael2k:
Design objection, and it supersedes part of my last review: **`HISTORY` should not ride the TNC control channel.** This belongs on the websocket.

The control channel is a line-oriented command/status protocol — `PTT`, `BUFFER`, `CONNECTED`, `DISCONNECTED` — driven from the reactor thread, and it is deliberately lossy: `tcp_write()` drops a line rather than block, because dropping a status message beats stalling the reactor. Pushing the persisted history through it fights that design at every step, and the follow-up commits show the cost:

- `tcp_write_all()` had to be generalised to the control port, changing the shared writer that every other caller depends on.
- The control port now accepts up to `WRITE_BACKPRESSURE_TIMEOUT_MS` (5 s) of reactor stall. A UI that connects and stops reading can delay `PTT`/`DISCONNECTED` handling for five seconds, where control writes previously never blocked. I flagged this as an acceptable trade; on reflection it is only acceptable because the transport is wrong.
- `HISTORY <n>` / `HISTORYMSG` / `HISTORYEND` needed hand-rolled framing, and the Go client needed the bulk channel sized from the header plus an "already in progress" re-entrancy guard — all of it re-implementing message boundaries and flow control that the websocket already provides.

The websocket is already the UI data path. `ws_broadcast_json()` exists, mongoose handles the draining and backpressure per connection (`is_draining`), and the Fyne UI already opens `ws://127.0.0.1:10000/websocket` — `link_engine_stub.go` says as much. Bulk, asynchronous, UI-facing data goes there; the chat history is exactly that.

What I would keep as-is: `common/message_store.{c,h}`, the four engine hooks, the `[store]` config, and the tests. That part is good work and is independent of transport — the module does not care who reads it.

What I would change:

- Drop the `HISTORY` TNC command and its `docs/TNC.md` section.
- Publish the history as JSON over the websocket when a UI client connects, using `msg_store_snapshot()` (which is already the right shape for this — one consistent buffer).
- Revert the `net.c` generalisation. `tcp_write_all()` goes back to being the DATA-stream writer and the control port keeps its non-blocking, drop-rather-than-stall behaviour.
- The Go client's history channel sizing, framing and re-entrancy guard all disappear with the command.

That also retires the first HIGH from my original review outright, rather than working around it.

The two findings from my previous comment still stand, since they are in the store module rather than the transport: the non-atomic file rewrite (`fopen(path, "w")` truncates the live file — use a temp file plus `rename()`), and that rewrite running inline on the ARQ event-loop thread.

rafael2k:
Re-reviewed `store-messages` @ 4124354. **The websocket move is right and my earlier objection is resolved** — history is off the TNC control port, `tcp_write_all` is back to the DATA-stream-only writer, and the control port keeps its drop-rather-than-stall behaviour. Net −246 lines in that commit, which is the shape a correct fix usually has.

Clean build here, 27/27 unit tests pass, 11/11 CI green.

I went after the new code specifically, since the refactor moved the framing out of `test_tcp_interfaces.c` (−146 lines of tests) and into `ui_communication.c`, where nothing covers it. Built a probe that drives the real store and runs the frame builder verbatim, under ASan+UBSan:

- 0, 1, 2, 3 and 4 messages → well-formed JSON every time, `{"type":"history","messages":[]}` for the empty store. No overflow, no sanitizer findings.
- Hostile content — embedded `"`, `\`, `,`, `[`, `]` in both peer and text — round-trips correctly and validates against a real JSON parser.
- The buffer sizing works out: prefix 30 + suffix 2 + at most `snap_len` of content against `snap_len + 64` allocated.

I also checked the case I expected to be a bug and it is not: `ws_broadcast_json` sends the history to **every** connected client whenever **any** new one connects, so an already-connected UI receives a second frame. That is harmless because the Go side does `state.history = e.Messages` — a replace, not an append — and the chat window is a singleton. Worth knowing it relies on that, though; an append there would duplicate the backlog.

Three things I would still change, none blocking:

**1. `if (snap[0]) free(snap);` decides ownership from content.** It is correct today only because `msg_store_snapshot()` documents that it returns NULL for an empty ring. If that ever returns an allocated empty buffer instead, this silently leaks on every UI connect. A separate `bool owned` (or just not aliasing `snap` to a string literal) says what is meant.

**2. The crash-safety comment claims more than the code delivers.** It says a crash "or power loss" leaves the old file intact, but the sequence is `fflush` → `fclose` → `rename` with no `fsync`. `rename` is atomic with respect to the directory entry; it does not make the new file's *data* durable. On ext4 with `data=ordered` this is usually fine thanks to the auto-fsync-on-rename heuristic, but that is a filesystem courtesy, not a guarantee. Either add `fsync(fileno(tmp))` before the `fclose`, or soften the comment to say what is actually guaranteed (crash-consistent, not power-safe). The store is chat history so the stakes are low — it is the claim I would fix, more than the behaviour. `fclose`'s return is also unchecked, though the preceding `fflush` check makes that mostly cosmetic.

**3. The frame builder deserves the test that moved away from it.** It is the one piece of new pointer arithmetic in the PR, it joins lines by scanning for `\n` while trusting `count` to match, and it is now the only untested step between the store and the UI. A small unit test over 0/1/N messages would have caught anything I had to write a throwaway probe to check.

Happy to see this merged once (1) and (2) are addressed — or as-is if you would rather land it and follow up, since none of the three is a correctness bug in the current code.

PR #257, opened 2026-09-07
arq: only send an ACCEPT in answer to a CALL we actually heard

rafael2k:
> **Correction, 2026-09-08.** The original description of this PR claimed these changes fix connect failures at the fringe, and cited a measured 4/20 baseline. **That evidence does not hold** and I have rewritten the description below. The code change is, I believe, still correct — but it should be judged as a protocol invariant, not as a reliability fix. See "Why the original evidence was wrong".

## What this changes

Half-duplex: the caller is deaf while it transmits. An ACCEPT is therefore only correct in the gap the caller opens by dropping PTT, and the only evidence we have that the gap is open is having just decoded a CALL.

Two paths in `ACCEPTING` did not respect that:

- **`ARQ_EV_RX_CALL`** reset the retry counter but left the deadline alone, so a retransmitted ACCEPT still fired relative to *our own* previous transmission rather than to the caller's. Now re-anchored to one channel guard after the heard CALL — the same schedule `fsm_listening` already uses for the first ACCEPT, and for the same reason.
- **`ARQ_EV_TX_COMPLETE`** arms the same `TIMER_RETRY` with `ARQ_ACCEPT_RX_WINDOW_MS` to hold the RX window open for the caller's first data burst. That is a *listening* window — but its expiry also transmitted an ACCEPT, one with no phase relationship to the caller at all. `accept_tx_pending` now records which of the two armed the pending timer; only a CALL-anchored one may key the transmitter. When the window expires we spend a retry slot and stay silent.

If the caller is still calling, we hear it and answer that. If we cannot hear it, transmitting blind only destroys someone else's burst.

This is arithmetic on protocol constants plus the CALL we decoded — **no carrier sense, no `channel_busy`**. It works against a hidden terminal, which sensing would not.

Unit test `test_accept_is_only_sent_in_answer_to_a_heard_call` pins the invariant: the CALL-anchored timer transmits, the window timer spends budget and stays silent, a fresh CALL re-arms both.

## Why the original evidence was wrong

The original description said two of three ACCEPTs in a failing connect were destroyed by the caller transmitting over them, and quoted a 4/20 connect-failure baseline.

The collision I traced was real. What was not real was the inference that it causes connect failures:

- Those runs were measured on the FIFO bridge, where **the same seed and binary gave 2995 / 3811 / 3523 blocks** across three consecutive runs. Any per-run comparison on that harness is noise. (Fixed since, in #259.)
- The A/B I ran gave **base 4/20 vs fix 5/20** — no improvement, well inside the noise for n=20.
- On the deterministic bench, **every ACCEPT in the failing runs goes out in clear air** with the caller listening. There is no collision to fix in the failures we actually have.

The real cause is that the handshake needs two marginal acquisitions to both succeed. Measured on the deterministic bench with a calibrated SNR axis, at −9.7 dB SNR3k:

| | acquired | delivered |
|---|---|---|
| DATAC16 standalone (`utils/acquire_vs_decode`, 100 bursts) | 80% | **61%** |
| Mercury, answerer decoding CALLs (10 windows) | 70% | **60%** |

Mercury matches the waveform's ceiling. Connect needs a CALL *and* an ACCEPT to survive at p≈0.6, which produces the ~20–25% failure rate we measure. It is DATAC16's cliff, not a bug.

## The second commit, with the same caveat

`tests/integration: drop audio the peer is not reading, like a radio does` — the FIFO bridge retried on `EAGAIN` until the peer drained, which makes the pipe a buffer holding a burst back through the peer's transmission and delivering it afterwards. A radio has no such buffer.

The principle is right. The specific measurement quoted in that commit message (an answerer raising `RX_CALL` 5.86 s late) came from overlaying two processes' traces that were **not on a common clock** — the trace stamps were process uptime, and the stations start at different times. Treat that number as unreliable; the argument stands on the mismatch with how a radio behaves, not on that figure.

## What I would do with this PR

Merge it if you agree the invariant is right — "never key over a burst you cannot hear, and only answer a CALL you actually heard" is correct regardless of whether it moves the failure rate. Do not merge it expecting better connect reliability at the fringe; that needs a more robust control exchange, and the deterministic bench can now measure such a change honestly.

### DIFF
--- a/datalink_arq/arq_fsm.c
+++ b/datalink_arq/arq_fsm.c
@@ -1343,7 +1343,6 @@ static void fsm_accepting(arq_session_t *sess, const arq_event_t *ev)
         sess->tx_retries_left   = ARQ_ACCEPT_RETRY_SLOTS;
         sess->accept_tx_pending = true;   /* answering a CALL we just heard */
         sess->deadline_ms       = time_now_ms() + ARQ_CHANNEL_GUARD_MS;
-        sess->deadline_event    = ARQ_EV_TIMER_RETRY;
         break;
 
     case ARQ_EV_TX_COMPLETE:
@@ -1361,11 +1360,11 @@ static void fsm_accepting(arq_session_t *sess, const arq_event_t *ev)
          * because that is the only moment we know the caller has dropped PTT
          * and is listening.  Firing one when this window expires has no phase
          * relationship to the caller at all, and lands in the middle of its
-         * next CALL, where a half-duplex radio is deaf.  Measured on the
-         * Watterson harness at SNR3k -9.8 dB: the answerer keyed an unprompted
-         * ACCEPT at +37.67 s while the caller transmitted CALL#4 from +34.55 to
-         * +38.27 s; two of the three ACCEPTs in that connect were destroyed
-         * that way and the call went unanswered.  4/20 connects failed.
+         * next CALL, where a half-duplex radio is deaf.
+         *
+         * This is a protocol invariant, not a reliability fix: the caller is
+         * deaf over its own transmission, so a blind ACCEPT is wrong whether
+         * or not it happens to collide in any given run.
          *
          * So mark the pending timer as "window", not "ACCEPT".  If the caller
          * is still calling we will hear it and answer that; if we cannot hear
--- a/datalink_arq/arq_fsm.c
+++ b/datalink_arq/arq_fsm.c
@@ -165,6 +165,10 @@ static void sess_enter(arq_session_t *sess, arq_conn_state_t new_state,
      * release. */
     if (new_state == ARQ_CONN_DISCONNECTED || new_state == ARQ_CONN_LISTENING)
         sess->deferred_listen_off = false;
+    /* The right to key an ACCEPT is earned by hearing a CALL, and it does not
+     * survive leaving ACCEPTING. */
+    if (new_state != ARQ_CONN_ACCEPTING)
+        sess->accept_tx_pending = false;
     if (new_state != ARQ_CONN_CONNECTED)
     {
         sess->pending_connect_confirm = false;
@@ -1088,6 +1092,7 @@ static void fsm_listening(arq_session_t *sess, const arq_event_t *ev)
         snprintf(sess->local_call, CALLSIGN_MAX_SIZE, "%s", ev->local_call);
         sess->session_id      = ev->session_id;
         sess->tx_retries_left = ARQ_ACCEPT_RETRY_SLOTS;
+        sess->accept_tx_pending = true;   /* answering a CALL we just heard */
         /* Reset mode state so the payload decoder matches the new caller's
          * initial DATAC15.  This must happen here (not in sess_enter for
          * DISCONNECTED/LISTENING) because LISTENING needs peer_tx_mode to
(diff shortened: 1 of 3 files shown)


rafael2k:
Pedro — review welcome, but please read the corrected description first: this PR is **not** in a state I would merge.

Short version: the change is a protocol invariant (an ACCEPT is only correct one channel guard after a CALL we actually heard, so the RX window timer must not transmit blind). The evidence originally offered for it was noise from a non-deterministic harness, and I have rewritten the description to say so.

What is now measured, on the deterministic bench that landed in #259:

- the collisions this fixes are real, but they are **not** what fails connects — in the failing runs every ACCEPT goes out in clear air
- connect failure at the fringe is DATAC16's own cliff: 61% delivery standalone at −9.7 dB, and mercury matches it at ~60%
- a long PTT tail (1.2 s) breaks the connect on **trunk** as well, so it is not a regression this introduces, and an ACCEPT-diversity variant I tried does not fix it either

So the open question is whether the invariant is worth landing on its own merits. I think it is, but I have no measurement showing it improves anything, and after today I would rather say that than imply otherwise.

pedromessetti (commented):
Reviewed the diff and ran the ARQ FSM suite (`make -C tests test_arq_fsm` → 38/38 pass). CI is green across all jobs.

**On the invariant itself: I agree it is correct on its own merits.** "Only key an ACCEPT one channel guard after a CALL you actually decoded" is a sound protocol invariant regardless of whether it moves the fringe connect rate, and the implementation reflects it cleanly. The `RX_CALL` re-anchor is a genuine fix: previously the retransmitted ACCEPT fired off the `TX_COMPLETE` deadline — a window sized for the caller's first DATA burst — which protected nothing and desynchronised the reply into the caller's next CALL. Mirroring `fsm_listening`'s one-guard schedule is the right move.

**`accept_tx_pending` lifecycle is handled correctly:** set on `RX_CALL` in both LISTENING and ACCEPTING, cleared by `sess_enter` when leaving ACCEPTING, cleared on `TX_COMPLETE` and on transmit. I checked the one plausible hazard — a late `TX_COMPLETE` clobbering a fresh `RX_CALL` re-anchor — and it is not reachable: half-duplex means we cannot decode a CALL while our own ACCEPT is still in flight, so `TX_COMPLETE` always precedes the next `RX_CALL`.

Items I'd want addressed before merge (none are functional):

1. **The discredited measurement is still asserted as fact in code comments.** The PR description (correctly) retracts the "Watterson harness … 4/20" evidence as noise, but three places still state it as fact:
   - `datalink_arq/arq_fsm.c:1364-1368` (TX_COMPLETE comment: "Measured on the Watterson harness … 4/20 connects failed")
   - `tests/datalink_arq/test_arq_fsm.c:977-980` (test doc: "…4 of 20 connect attempts failed outright")
   - `tests/integration/channel_bridge_test.go:400-402` ("Measured at SNR3k -9.8 dB: an answerer raised RX_CALL 5.86 s …")
   The 5.86 s figure is explicitly called unreliable in the description (trace clocks not on a common clock). Please rewrite these to the invariant framing so future readers don't re-cite a measurement you've shown does not hold.

2. **`delivered` diagnostic now over-counts.** In `channel_bridge_test.go` `outBytes` is still `+= whole*2` unconditionally (line 422) even when the EAGAIN branch dropped the tail, so "delivered" is really "presented to the FIFO" and `delivered + dropped > keyed`. Either subtract the dropped tail from `outBytes`, or rename the CHDIAG field to `presented`.

3. Minor: `sess->deadline_event = ARQ_EV_TIMER_RETRY` in the ACCEPTING `RX_CALL` handler (`arq_fsm.c:1346`) is redundant — the deadline event is already `TIMER_RETRY` throughout ACCEPTING. Harmless either way.

4. Minor: `tx_retries_left` now counts "slots (transmit or silent wait)" rather than "ACCEPT transmissions". This changes the give-up duration (faster now — a pending call no longer sits through four ACCEPT+RX-window cycles), which is arguably an improvement, but the field name/comment (`retries remaining for current frame`) no longer matches its ACCEPTING semantics. A one-line note would prevent future confusion.

The new unit test `test_accept_is_only_sent_in_answer_to_a_heard_call` pins the right thing and is well constructed. Optional additions: assert the re-anchored `deadline_ms` value after `RX_CALL`, and cover the full give-up-to-LISTENING path under the silent-wait semantics (currently only the pre-existing tests exercise give-up, implicitly under the new behavior).

The integration harness change (drop-on-EAGAIN instead of retry) is the correct fidelity fix for a half-duplex peer; the write loop logic (partial write, then drop remainder on EAGAIN) is right given `rxFD` is `O_NONBLOCK`.

rafael2k:
**Scope:** the harness change is gone — this PR is now three files: `arq_fsm.c`, `arq_fsm.h`, `test_arq_fsm.c`. The channel-bridge realism fix and Pedro's CHDIAG accounting fix moved to #267, which they never had anything to do with beyond being written the same week. Combined, the two branches are byte-identical to what this PR contained before the split.

**And the reason this PR exists is now measured, not argued.**

I owed this. The original justification was a harness measurement I later showed was invalid, and after that the PR rested on a protocol argument alone. So I went looking for the failure it claims to fix, expecting not to find it — and at first I didn't. Eleven runs across clean and fringe SNR: zero collisions, both arms behaving identically, no difference in anything. I was ready to close this.

The regime was just narrow. The answerer has to *hear* a CALL (so it sends an ACCEPT) while the caller *misses* the ACCEPT (so the timer path engages). Scanning the noise axis finds it at `MERCURY_CH_NO=-3`:

| | connects | collisions | doubled airtime |
|---|---|---|---|
| trunk | **0 / 3** | **5** | **10.43 s** |
| this PR | **2 / 3** | **0** | **0.00 s** |

A "collision" here is both stations keying at the same instant, taken straight from the PTT logs.

The mechanism is visible in one run. The answerer keys at 28.4, 41.1, 53.8, 66.6 s — a fixed ~12.7 s timer, firing blind. The caller keys at 32.9, 37.6, 53.0, 68.4, 83.8 s — its ~15.4 s call interval. Two independent periods, beating against each other, overlapping at 41.10–41.93, 53.84–57.35 and 68.39–70.28.

The consequence is the interesting part. In that run the answerer keyed **4 times total** and the caller **26 times**: every blind ACCEPT landed on the caller's own transmission, where a half-duplex radio is deaf, so the caller never heard one and kept calling until it gave up. With this change the answerer keys 8–9 times, each one an answer to a CALL it actually heard, and the connect completes.

That is exactly what the invariant predicts: an ACCEPT is only correct one channel guard after a heard CALL, because that is the only moment the caller is known to have dropped PTT. A timer has no phase relationship to the caller at all.

**Caveats, stated plainly.** Three runs per arm, one noise point, on the FIFO bridge harness. `No=-3` is a narrow band — at `-4` and cleaner both arms connect on the first exchange and this change does nothing; at `+2` neither connects because the answerer never hears a CALL at all. The third run of each arm failed to connect on both sides, so the connect improvement is 2/3 vs 0/3 on a small sample. What I would not claim from this is a throughput number. What it does establish is that the collision is real, reproducible, and absent with the change.

Build clean, 28/28 unit tests, rebased on current trunk.

PR #261, opened 2026-09-08
build: track header dependencies, so a header edit rebuilds what includes it

rafael2k:
## The bug

The build had **no header dependency tracking at all**. No Makefile passed `-MMD`, and none `-include`d a `.d` file. So editing a header rebuilt nothing that included it, and the link mixed objects compiled against different versions of the same struct.

## Why it matters, concretely

Adding a field to `arq_session_t` in `arq_fsm.h` and rebuilding incrementally produced a `mercury` whose ARQ event loop read `deadline_ms` at the wrong offset. It sent one CALL and then sat forever — with a live event loop parked in `pthread_cond_timedwait`, so it looked like a perfectly healthy process that had simply stopped retrying.

That was chased as a protocol bug: through the FSM, through gdb thread dumps, through the harness. The build was the last thing suspected.

Worse, **two separate experiments produced confident and wrong conclusions** from binaries built that way, and both had to be thrown out:

- "a 1.2 s PTT tail is a regression introduced by PR #257" — it is not; clean builds show trunk failing identically
- "ACCEPT diversity fixes the long-tail case" — it does not

Neither error was visible at the time. That is the real cost: not the hour lost, but that the build silently manufactures evidence.

## The fix

- `-MMD -MP` in `COMMON_CFLAGS` (config.mk)
- each Makefile `-include`s the `.d` files beside **its own** objects — the paths inside a `.d` are relative to the directory the compiler ran in, so this cannot be done once at the top level
- `-MP` emits a phony target per header, so deleting or renaming one does not wedge the build with *"no rule to make target"*

(`*.d` was already in `.gitignore` since #199, and the stale `.d` files left by that older build were already untracked/absent on disk, so this PR adds no `.gitignore` entry and deletes nothing — it only introduces the `-MMD -MP` generation and the per-Makefile `-include` consumption.)

## Verification

```
$ touch datalink_arq/arq_fsm.h && make
   → objects that include arq_fsm.h rebuilt      (before this change: 0)
```

Clean build green, unit suite green, `TestMercuryARQTransfer` green.

Worth watching on this PR: the cross-compile jobs (Windows/arm64/armhf) and the macOS job, since `-MMD -MP` now applies to every compiler in the matrix.


rafael2k:
Pedro — review welcome on this one especially, since it changes the build for every target.

The substance is small (`-MMD -MP`, and each Makefile `-include`ing the `.d` files beside its own objects), but two things are worth a second pair of eyes:

- Each Makefile has to include its own `.d` files, because the paths inside a `.d` are relative to the directory the compiler ran in. If you see a subdirectory I have missed, that directory silently keeps the old behaviour.
- `-MMD -MP` now applies to every compiler in the matrix, including the mingw and cross-compile paths. CI is green on all of them, but you know the Windows build better than I do.

Context for why it matters: before this, editing a header rebuilt nothing that included it. Adding a field to `arq_session_t` and rebuilding incrementally gave a binary whose ARQ event loop read `deadline_ms` at the wrong offset — it sent one CALL and then sat forever with a healthy-looking event loop. I chased that as a protocol bug through gdb and the FSM before suspecting the build, and two experiments reached confident wrong conclusions from binaries built that way.

Makefile
```diff
@@ -867,3 +867,8 @@ test:
 
 integration-test:
 	cd tests/integration && go test -v ./...
+
+# Pull in the header dependencies generated by -MMD (see config.mk).  Without
+# this, a header edit does not rebuild the objects that include it and the link
+# silently mixes old and new struct layouts.
+-include $(wildcard *.d */*.d */*/*.d */*/*/*.d)
```
pedromessetti: This wildcard only reaches 4 path segments, but the RaptorQ `deps/obl` objects built by this same Makefile (`libmercury_core.a` → `$(RAPTORQ_OBJS)`) land at 5 segments (`datalink_broadcast/raptorq/deps/obl/oblas_*.d`), so they get no header tracking. Also, `*/*.d */*/*.d */*/*/*.d` sweeps in `.d` files generated by the sub-makes (common/, datalink_arq/, …) whose paths are relative to those subdirectories; resolving them here is wrong (harmless today only because those targets are never built at the top level).

pedromessetti (commented):
Thanks — the root cause is real and the `-MMD -MP` + per-Makefile `-include` approach is the right shape. I verified locally that touching `datalink_arq/arq_fsm.h` now rebuilds the dependent objects (`arq.c`, `arq_fsm.c`, `arq_modem.c`, `modem.c`) where before it rebuilt none.

A few issues, ordered by importance:

**1. The `.d` wildcard depth misses the RaptorQ `deps/obl` objects.** `datalink_broadcast/raptorq/deps/obl/oblas_*.c` are compiled by the *top-level* Makefile (via `libmercury_core.a` → `$(RAPTORQ_OBJS)`), so their `.d` files land at 5 path segments (`datalink_broadcast/raptorq/deps/obl/oblas_lite.d`). `$(wildcard *.d */*.d */*/*.d */*/*/*.d)` only matches 4, so those objects get no header tracking at all. Consider adding a `*/*/*/*/*.d` level, or deriving the `.d` list from the object list instead of a fixed depth.

**2. `modem/freedv/Makefile` was not updated.** The codec2 objects under `modem/freedv/` generate `.d` files, but the nested `make -C freedv` Makefile has no `-include`, so header edits there do not rebuild. The parent `modem/Makefile` `*/*.d` wildcard cannot cover it correctly because those `.d` paths are relative to `modem/freedv/`.

**3. The broad wildcard cross-includes `.d` files owned by other directories.** The `config.mk` comment correctly says paths in a `.d` are relative to the directory the compiler ran in "so they must be included from there" — yet `-include $(wildcard *.d */*.d */*/*.d */*/*/*.d)` in each Makefile (and the top level) sweeps in sibling/subdirectory `.d` files whose relative paths then resolve against the wrong directory. It is harmless today only because those `.d` files define targets the including Makefile never builds (orphans), but it creates misleading dependency edges and bogus `-MP` phony header targets. Safer to include only the `.d` files beside the objects each Makefile actually owns.

**4. `clean` does not remove `.d` files (except `datalink_broadcast`).** `make clean` left 114 `.d` files behind. `-MP` avoids the classic "no rule to make target" wedge, but leaving stale `.d` after clean is inconsistent — `datalink_broadcast/Makefile` already cleans its own. Recommend adding `*.d` to the other clean targets.

**5. The hidapi object rules do not pass `$(CFLAGS)`.** `radio_io/hidapi-w64/hid.o` and `radio_io/hidapi-macos/hid.o` (Makefile:236–242) compile with a bare `-O2 -I…`, so they get no `-MMD -MP`. Low impact (vendored, rarely edited), noting for completeness.

**6. The PR description overstates what the diff does.** The body says it "deletes the 153 stale `.d` files already on disk" and "adds `*.d` to `.gitignore`", but the diff does neither — `.gitignore` already had `*.d` (since #199), and no `.d` files are deleted here. The `.d` files were already untracked/absent; only the new `-MMD -MP` + `-include` behavior is new. Worth tightening the description for an accurate record.

modem/Makefile
```diff
@@ -21,3 +21,8 @@ framer.o: framer.c framer.h
 clean:
 	$(MAKE) -C freedv clean
 	rm -f *.o
+
+# Pull in the header dependencies generated by -MMD (see config.mk).  Without
+# this, a header edit does not rebuild the objects that include it and the link
+# silently mixes old and new struct layouts.
+-include $(wildcard *.d */*.d */*/*.d */*/*/*.d)
```
pedromessetti: `make -C freedv` builds the codec2 objects in `modem/freedv/`, whose `.d` files have paths relative to that directory. Those `.d` files are never consumed — `modem/freedv/Makefile` has no `-include`, and this wildcard matching `freedv/*.d` resolves those relative paths against `modem/`, so freedv header edits do not trigger rebuilds.

Issue #254, opened 2026-09-05
macOS Tahoe 26.6.2: CoreAudio capture fails after opening successfully


N4EAC:
## macOS Tahoe 26.6.2: CoreAudio capture fails after opening successfully

Hello Rafael,

During the initial radio tests you requested for the recently merged Mercury changes, we discovered what appears to be a separate CoreAudio capture issue on macOS Tahoe.

The configured radio frequency is read successfully through Hamlib, Mercury starts normally, and the UI WebSocket connects. Mercury also enumerates and opens the selected USB audio device. However, CoreAudio capture never delivers samples. Mercury subsequently detects a stalled capture stream, and its attempt to reopen the device fails continuously.

### Environment

- macOS Tahoe 26.6.2
- macOS build: `25G83`
- Architecture: Apple Silicon `arm64`
- Mercury: `1.9.13`
- Mercury revision: `7febb890`
- Binary: native `Mach-O 64-bit executable arm64`
- Hamlib runtime: `4.7.2`
- Radio: Yaesu FT-710
- Audio device: radio USB audio interface, displayed by CoreAudio as `USB Audio Device`
- Audio subsystem: `coreaudio`
- Selected native CoreAudio device ID during this run: `124`
- Capture channel: left
- TX gain: `0.00 dB`

Mercury was running as the bundled subprocess of Mercury SkyPulse 0.1.7.

### Configuration

Relevant configuration:

```ini
[main]
input_device = "124"
output_device = "124"
capture_channel = left
sound_system = coreaudio

[audio]
tx_gain_db = 0.00
tx_delay_ms = 10
```

The device ID was selected from the device list reported by Mercury. It was not manually guessed or derived from an operating-system device name.

### Steps to reproduce

1. Connect the FT-710 USB audio and CAT interfaces to an Apple Silicon Mac.
2. Start Mercury with `sound_system = coreaudio`.
3. Select the Mercury-enumerated USB audio device for both capture and playback.
4. Confirm macOS microphone/audio-input authorization.
5. Start Mercury and observe the capture and WebSocket logs.
6. Supply receive audio from the radio.

### Observed result

Mercury initially reports successful playback and capture initialization:

```text
[INF] [engine] Initialising audio I/O (subsystem=4)
[INF] [radio-io] Initializing PTT (method=hamlib, device=<redacted>)
[INF] [radio-io] Hamlib serial speed overridden to 115200 baud
[...]

PR #264, opened 2026-09-08
audioio: follow a device that comes back under a new id

rafael2k:
> **Scope correction.** This fixes re-enumeration **within a running process** only. It does **not** fix the across-restarts case, which is probably the more common one. See "What this does not fix".

Addresses part of the reopen half of #254.

## The mechanism

CoreAudio assigns `AudioDeviceID`s per enumeration, so a USB interface that drops off the bus and comes back gets a **new** one. The UI sends the numeric id and it is stored verbatim (`input_device = "124"`), so once the id moves it points at nothing and the open fails inside `AudioObjectGetPropertyData`.

**Verified on macOS 15.7.9**: an id absent from the device list fails exactly that call, reported by the new diagnostics as `'!obj'` (`kAudioHardwareBadObjectError`, 560947818) — the same call whose bare name appears in #254's reopen loop.

## What this fixes

Within one process: record the name each direction resolved to, and if that id later stops appearing in the device list, follow the name to whatever id it holds now. Only on an unambiguous match, since duplicate display names are common and binding the wrong one would key the wrong radio.

That covers "the device glitched mid-session and came back", which is the shape of #254's stall-then-reopen loop.

## What this does NOT fix

The name is held in a process-local static. On a **fresh start** it is empty, and the config still holds a numeric id. So if the device was re-enumerated while Mercury was not running — unplug the radio, reboot, start Mercury — the id is stale, there is nothing to fall back to, and the open fails exactly as before.

That case needs a **stable identifier persisted in the config**, not a volatile index. The resolver already accepts a name, so the missing piece is on the writing side. Sketch:

- persist the device **name** alongside the id (e.g. `input_device_name`), written when a device is selected or successfully opened
- at open: use the id if it still resolves *and* still carries that name; otherwise re-resolve by name
- keep the id as a fast path and a tie-break for duplicate names

That is a config-schema and UI change, so it is deliberately not in this PR.

## Risk

Narrow. On the normal path nothing changes but recording a name. The new branch is reached only where the old code was already passing through an id that could not resolve.

## Tested

macOS 15.7.9 (x86_64 VM, BlackHole 2ch): open by id, open by name, capture runs, clean shutdown; bogus id fails with `'!obj'`. Linux and CI green including macOS arm64.

**Not tested: the fallback branch itself** — it needs a device to genuinely disappear between opens, which that VM cannot stage. It wants the reporter's hardware or a USB device unplugged mid-run.

### DIFF
--- a/audioio/audioio.c
+++ b/audioio/audioio.c
@@ -1987,9 +1987,27 @@ int get_soundcard_list(int audio_system, int mode,
  * enumerable devices (NULL/FIFO/SOCK/SHM), enumeration fails, or a
  * substring match is ambiguous, buf is left untouched -- callers get
  * exactly today's behavior. */
+/* Display name of the device each direction last resolved to.
+ *
+ * A native device id is not necessarily stable: CoreAudio AudioDeviceIDs are
+ * assigned per enumeration and a USB interface that drops off the bus and
+ * comes back gets a NEW one.  A config holding a numeric id (which is what the
+ * UI writes) then points at nothing, and every reopen fails in
+ * AudioObjectGetPropertyData -- reported, before the diagnostics landed, as a
+ * bare "AudioStreamBasicDescription" and, with them, as '!obj'
+ * (kAudioHardwareBadObjectError).  Verified on macOS: an id that is not in the
+ * device list produces exactly that.
+ *
+ * Names survive re-enumeration where ids do not, so remember the name we
+ * opened and fall back to it when the id stops resolving.  This is the reopen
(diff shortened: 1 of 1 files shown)


rafael2k:
Superseded by the UID work: identifying CoreAudio devices by `kAudioDevicePropertyDeviceUID` resolves the device on every open, so it covers the within-session re-enumeration this PR handled *and* the across-restart case this PR did not. Nothing here needed carrying over — the process-local name cache was a workaround for a stored identifier that should never have been an enumeration index.

Replacement PR follows.

PR #268, opened 2026-09-10
tcp: make broadcast listener shutdown bounded

N4EAC:
The broadcast server could keep shutdown blocked in two independent waits: `accept()` with no client, and a sender stranded in `read_buffer()` when disconnect teardown cleared its ring before joining it.

This makes the listener and client sockets nonblocking with bounded polling, makes send and receive loops observe shutdown, and joins the sole ring reader before clearing its buffer. It also handles worker creation failures without joining an uninitialized thread. The existing message-store behavior and current broadcast framing changes are preserved.

Hardware-free regressions use real loopback sockets and the real ring implementation to cover:

- no connected client
- an idle connected client
- a client that has disconnected
- a backpressured send
- disconnect during the sender's size/read race

Validation:

- Current `mercuryv2` base `8cbdb9c`, macOS: 73/73 focused tests
- Current base, macOS ASan/UBSan: 73/73 focused tests
- Earlier identical shutdown correction, Windows 11: 20/20 no-client and 3/3 idle-client production shutdowns, 140–407 ms
- Earlier identical shutdown correction, Ubuntu 24.04 VMware: 20/20 no-client and 3/3 idle-client production shutdowns, 115–119 ms, plus focused native and sanitizer passes
- FT-710 receive-only validation on unmodified trunk reproduced the separate listener hang after both audio threads exited; no CAT/PTT or RF transmission was used

### DIFF
--- a/data_interfaces/tcp_interfaces.c
+++ b/data_interfaces/tcp_interfaces.c
@@ -120,17 +120,27 @@ static ssize_t send_all(int socket_fd, const uint8_t *buffer, size_t len)
 {
     size_t total_sent = 0;
 
-    while (total_sent < len)
+    while (total_sent < len && !shutdown_ && !bcast_client_done)
     {
         ssize_t sent = send(socket_fd, (const char *)buffer + total_sent, len - total_sent, HERMES_SEND_FLAGS);
-        if (sent <= 0)
+        if (sent < 0)
         {
+            int err = sock_errno();
+            if (err == SOCK_EINTR)
+                continue;
+            if (err == SOCK_EAGAIN || err == SOCK_EWOULDBLOCK)
+            {
+                hermes_usleep(20000);
+                continue;
+            }
             return -1;
         }
+        if (sent == 0)
+            return -1;
         total_sent += (size_t)sent;
     }
 
-    return (ssize_t)total_sent;
+    return total_sent == len ? (ssize_t)total_sent : -1;
 }
 
 typedef struct
@@ -1239,10 +1249,9 @@ void *send_thread(void *client_socket_ptr)
 
     while (!shutdown_ && !bcast_client_done)
     {
-        /* Poll size_buffer() instead of blocking in read_buffer()'s
-         * condvar wait.  Re-check bcast_client_done after size_buffer()
-         * returns to close the race where clear_buffer() drains the
-         * ring between our size check and the read_buffer() call. */
+        /* Only this worker consumes the RX ring, and the server clears it
+         * after joining us. A full frame therefore cannot disappear between
+         * this size check and read_buffer(). Poll while no frame is ready. */
         if (size_buffer(data_rx_buffer_broadcast) < frame_size)
         {
             msleep(100);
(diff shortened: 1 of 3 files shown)


rafael2k:
Reproduced the hang and confirmed the fix independently on macOS Tahoe 26.6.2 (25G83), VM with a VoodooHDA capture/playback device. Broadcast port open, no client connected, SIGTERM, three runs each:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| trunk (`aa783af`) | 10.13 s | 10.10 s | 10.12 s |
| this PR (`8afe671`) | **0.21 s** | **0.21 s** | **0.21 s** |

The detail that matters is in the shell's report on every trunk run:

```
Alarm clock: 14   ./mercury -v -x coreaudio ...
```

Trunk does not shut down at all — it is **SIGALRM'd by the `alarm(10)` safety net**. So the listener hang is real, the watchdog is the only thing ending the process, and this PR is what lets it exit on its own. 0.21 s, identical across runs.

`test_tcp_interfaces`: 73/73 on macOS, matching your figure.

On the change itself, the part I looked at hardest is the teardown ordering, since that area has bitten us before — the earlier UAF was rings freed under live modem threads, fixed by an order swap plus this same `alarm(10)`. Joining the sole ring reader *before* clearing its buffer is the right shape: it removes the size-check/`read_buffer()` race by ordering instead of papering over it with poll-and-recheck, and the comment now states the invariant that makes it safe ("only this worker consumes the RX ring, and the server clears it after joining us") rather than describing the race it used to dodge. Handling `pthread_create` failure without joining an uninitialised thread is a real bug caught in passing.

Reads good to me and the evidence holds up on a third platform. Merging unless you would rather I wait.

N4EAC:
> Reproduced the hang and confirmed the fix independently on macOS Tahoe 26.6.2 (25G83), VM with a VoodooHDA capture/playback device. Broadcast port open, no client connected, SIGTERM, three runs each:
> 	run 1 	run 2 	run 3
> trunk (`aa783af`) 	10.13 s 	10.10 s 	10.12 s
> this PR (`8afe671`) 	**0.21 s** 	**0.21 s** 	**0.21 s**
> 
> The detail that matters is in the shell's report on every trunk run:
> 
> ```
> Alarm clock: 14   ./mercury -v -x coreaudio ...
> ```
> 
> Trunk does not shut down at all — it is **SIGALRM'd by the `alarm(10)` safety net**. So the listener hang is real, the watchdog is the only thing ending the process, and this PR is what lets it exit on its own. 0.21 s, identical across runs.
> 
> `test_tcp_interfaces`: 73/73 on macOS, matching your figure.
> 
> On the change itself, the part I looked at hardest is the teardown ordering, since that area has bitten us before — the earlier UAF was rings freed under live modem threads, fixed by an order swap plus this same `alarm(10)`. Joining the sole ring reader _before_ clearing its buffer is the right shape: it removes the size-check/`read_buffer()` race by ordering instead of papering over it with poll-and-recheck, and the comment now states the invariant that makes it safe ("only this worker consumes the RX ring, and the server clears it after joining us") rather than describing the race it used to dodge. Handling `pthread_create` failure without joining an uninitialised thread is a real bug caught in passing.
> 
> Reads good to me and the evidence holds up on a third platform. Merging unless you would rather I wait.

Good morning Rafael,
Thank you for reproducing this independently and reviewing the teardown ordering. Your results match ours, and I have no reason to delay. Please proceed with the merge.

PR #269, opened 2026-09-10
broadcast: fix the RaptorQ symbol size so symbols survive a mode change

rafael2k:
**Wire-format change, and it drops two modes from broadcast.** Both flagged up front — that is why this is a PR.

*(This briefly sat on trunk by my mistake and was reverted in aa783af; this is the same work, unchanged, put up for review.)*

### The problem

`symbol_size = frame_size - overhead`, so T is a function of the MODE. That is the most efficient possible packing and it has one fatal consequence: symbols collected at one mode are worthless at another. A mode change mid-transfer throws away everything the receiver has accumulated, and one carousel cannot serve a fast audience and a fringe audience at the same time.

### The change

T fixed at 41 bytes; a frame carries as many whole symbols as it holds:

| mode | frame | symbols | payload efficiency |
|---|---|---|---|
| QAM16C2 | 1213 | 29 | 98.0 % |
| DATAC17 | 1180 | 28 | 97.3 % |
| DATAC1 | 510 | 12 | 96.5 % |
| DATAC3 | 126 | 2 | 65.1 % |
| DATAC4 | 54 | 1 | 75.9 % |

The symbols in a frame are **consecutive ESIs of one block**, so the frame's single existing tag describes all of them — the receiver reads the base sbn/ESI and counts. Tagging each symbol separately would cost 3 bytes apiece (6.8 points at QAM16C2) to buy a per-symbol block spread a fountain code does not need, since every block must reach K+ε either way.

**Cost of mode-independence: one point.** 98.0 % against 99.0 % for the old mode-locked packing — against the 100 % extra a second carousel for the fringe audience would cost.

### What it buys

- A mode change costs nothing. With T fixed the OTI no longer depends on the mode, so symbols already received stay valid and the receiver need not reset at all.
- One carousel can interleave modes — fast for good signal, slow for fringe, every symbol counting toward the same object. The receiver derives the symbol count from **the length of the frame it decoded**, not from any configured mode, so it needs no telling which mode is on the air. Not wired into the sender yet; the architecture is ready.

### What it costs — please review this specifically

**FSK_LDPC and DATAC15 leave broadcast.** Both are 30-byte frames: they held an 18-byte symbol under the old sizing and cannot hold a 41-byte one. Keeping them would mean T ≤ 17 and 84 % efficiency at the fast end. DATAC16 and DATAC0 (14 B) could never have participated. The refusal message now states the actual shortfall, and a test pins exactly which modes are usable so this cannot drift silently.

### Also here

A running decode is reset when the **OTI changes**, not only when the session id does — a session id is a random 1..31 and collides one time in 31, and without this, symbols describing a different object would be fed to a decoder built for the old OTI. And a frame's ESI run may not straddle the 16-bit ceiling, since one tag can only describe a base; the run is shortened instead of wrapping.

### Tests

- `test_symbol_size_is_fixed_across_modes` pins the symbol count for every mode, and therefore the cost. If T ever reverts to frame-derived, every property above dies quietly and this is what notices.
- `test_symbols_from_two_modes_decode_one_object` feeds a receiver alternate QAM16C2 and DATAC4 frames, dropping half of each, and requires completion — neither sender's frames alone suffice. Interleaving already works at the codec level.

Two existing tests needed honest updates rather than accommodation: the "small robust mode" case moves from DATAC16 to DATAC4 (the new floor), and the raw-payload case imitates a foreign sender at the packed layout instead of one-symbol-per-frame.

19/19 broadcast tests, 28/28 unit tests, clean build.

### DIFF
--- a/datalink_broadcast/bcast_file.c
+++ b/datalink_broadcast/bcast_file.c
@@ -505,7 +505,13 @@ bcast_rx_result_t bcast_file_rx_frame(bcast_file_rx_t *rx,
 
     /* Anything that is not exactly one of our frames is somebody else's
      * traffic on the broadcast plane -- chat, KISS from another client -- and
-     * must be passed over silently rather than treated as corruption. */
+     * must be passed over silently rather than treated as corruption.
+     *
+     * This is also the gate that confines a receive to ONE mode.  Relaxing it
+     * to "any usable mode's frame size" is this side's half of interleaving;
+     * the other half is modem.c -- its single broadcast_frame_size filter, and
+     * pointing the two rx workers of the dual receiver at the two interleaved
+     * modes.  See bcast_modes.h. */
     if (len != rx->frame_size) return BCAST_RX_IGNORED;
     if (((frame[0] >> BCAST_PACKET_TYPE_SHIFT) & BCAST_PACKET_TYPE_MASK)
             != BCAST_PACKET_RQ_CONFIG)
@@ -523,19 +529,19 @@ bcast_rx_result_t bcast_file_rx_frame(bcast_file_rx_t *rx,
      * message (chat would never see it) and reset a decode in progress.
      *
      * So check that the frame's own OTI describes a transfer that could
-     * actually be arriving on THIS mode: the declared symbol size must match
-     * our frame size exactly, and the declared object length must be sane.
-     * Text would have to encode T-1 in two specific bytes by accident, which
-     * is a 1-in-65536 coincidence on top of the header, and then survive the
-     * length check as well. */
+     * actually be one of ours: the declared symbol size must be exactly our
+     * fixed T, and the declared object length must be sane.  Text would have
+     * to encode T-1 in two specific bytes by accident, which is a 1-in-65536
+     * coincidence on top of the header, and then survive the length check as
+     * well. */
     {
         uint32_t f_len = (uint32_t)frame[1] | ((uint32_t)frame[2] << 8) |
                          ((uint32_t)frame[3] << 16);
         uint32_t t_dec = ((uint32_t)frame[4] | ((uint32_t)frame[5] << 8)) + 1;
-        /* The symbol size is FIXED and mode-independent, so this check no
-         * longer says "did this arrive on my mode" -- it says "is this one of
-         * ours at all".  That is what allows an interleaved carousel: frames
-         * from any mode carry the same T and are all acceptable. */
+        /* T is fixed now, so this no longer says "did this arrive on my
+         * mode" -- it says "is this one of ours at all".  Mode-matching is
+         * still enforced, by the frame-size gate at the top of this function;
+         * see bcast_modes.h on why interleaving is not yet reachable. */
         if (t_dec != (uint32_t)rx->symbol_size)
             return BCAST_RX_IGNORED;
         if (f_len == 0 || f_len > BCAST_FILE_MAX_BYTES)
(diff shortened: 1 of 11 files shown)


rafael2k:
**This applies to the Fyne UI file transfer too**, which I had not checked when I opened this. It goes through the same `bcast_file` API via `mercury_bridge.c`, so the wire change and the mode removal both reach it — and the UI was not equipped to say so.

`mercury_bcast_engine_mode()` returned −1 only when the running FreeDV mode was absent from its table. It never consulted `bcast_file_mode_usable()`, so with the engine on FSK_LDPC or DATAC15 it returned a valid index and every caller proceeded as though broadcast would work, until `tx_open` failed and handed the operator a raw C error string. Receiving failed the same way.

Meanwhile `broadcastModeUsable()` — the exact predicate needed — was defined in Go and **called from nowhere**. That turns out not to be an oversight but a consequence: until the symbol size was fixed, every mode the engine could run either worked for broadcast or was missing from the table, so "runnable" and "usable" never disagreed. FSK_LDPC and DATAC15 are the first modes for which those two answers differ, which is exactly what makes this PR surface the gap.

Fixed in the bridge rather than the UI, so every consumer gets it: `engine_mode()` now reports −1 for a runnable-but-unusable mode, which makes the existing "this modem's mode cannot carry broadcast" guards fire *before* a transfer is attempted. Those guards were already written and already correct — they simply never ran.

`engine_mode_raw()` is added alongside so the refusal can name the mode:

> Mode: DATAC15 cannot carry broadcast - its frames are too small to hold one RaptorQ symbol. Restart mercury with -m 3 (DATAC4) or faster.

which tells an operator what to change, where "this modem's mode" sends them hunting. The hermes-to-FreeDV table is hoisted to file scope so the two accessors share one definition rather than drifting apart.

New Go test pins usability for all eleven modes and cross-checks each verdict against the frame geometry, so the predicate and the 41-byte floor cannot disagree silently.

`mercury` and `fyne-ui` both build; 28/28 unit tests; fyne-ui suite passes.

rafael2k:
Added the operator-facing documentation (`de5510e`), and verifying it end to end caught a mistake worth recording.

## Sending an NNCP bundle

```
# --- receiving station ---
mercury -x alsa -i plughw:1,0 -o plughw:1,0 -m 10 -b 8100 &
bcast_file_tool recv /var/spool/nncp/incoming -m 10 -p 8100

# --- sending station ---
nncp-bundle -tx somenode > /tmp/out.nncp
mercury -x alsa -i plughw:1,0 -o plughw:1,0 -m 10 -b 8100 &
bcast_file_tool send /tmp/out.nncp -m 10 -c 12 -p 8100
```

The file arrives under its **original name** (the name travels in the bundle), so NNCP's inbound spool picks it up directly. Mercury's "bundle" is just file + name and is unrelated to NNCP's own bundle format, which it treats as opaque bytes.

**The mode must match in four places** — both Mercury instances and both tool invocations. There is no negotiation in broadcast and `-m` cannot be changed at runtime. A mismatch produces *no error*, because a frame on the wrong mode is indistinguishable from noise. That is the first thing that bites, so the doc now leads with it.

## The cycle count, which this PR changes

My first attempt at the example used `-c 2` and **delivered nothing**. That is not a bug — it is the arithmetic changing under the fixed symbol size, and it is exactly the stale instinct the doc needs to head off:

```
symbols needed  ~=  bundle_bytes / 41     (plus a couple for RaptorQ)
frames needed   ~=  symbols / symbols_per_frame
```

A 6 kB bundle on QAM16C2 needs 147 symbols at 29 per frame = **6 frames**. Under the old mode-sized symbol (1200 B) the same file was ~6 symbols and 2 cycles was ample. Now `-c 2` sends 58 symbols and the receiver just logs progress and never completes.

Verified over `ch`:

```
sending /tmp/out.nncp: bundle 6016 B, mode 10 (1213 B/frame), 1 block(s), cycles 12
  receiver: symbols 29 -> 58 -> 87 -> 116 -> 145
  received "out.nncp" -> .../inbox/out.nncp
  RESULT: byte-identical (6000 B)
```

Completed on the 6th frame, as predicted. On a fading link the receiver needs the same 147 *distinct* symbols, so budget for loss (30 % → ~1.4× frames); `-c 0` repeats until interrupted, which is the honest default with no return path.

## Mode table corrected for the fixed symbol

The doc's old "small-frame modes are not usable" section was stale. It now gives symbols-per-frame and states the loss plainly:

| mode | frame | symbols/frame | file transfer |
|---|---|---|---|
| QAM16C2 | 1213 | 29 | yes |
| DATAC17 | 1180 | 28 | yes |
| DATAC1 | 510 | 12 | yes |
| DATAC3 | 126 | 2 | yes — sensible robust choice |
| DATAC4 | 54 | 1 | yes — the floor |
| FSK_LDPC, DATAC15 | 30 | 0 | **no — regression** |
| DATAC0/13/16 | 14 | 0 | no (never could) |
| DATAC14 | 3 | 0 | no |

And what it does *not* affect: **chat and GPS still work on DATAC15 and FSK_LDPC** — a chat line only needs a frame, and DATAC15 still carries 18 characters. The 41-byte floor is a file-transfer constraint only.

PR #285, opened 2026-09-15
ci: pin the certum-signer image to the digest that signed 1.9.13

rafael2k:
Unblocks the **1.9.14 release**.

## What happened

reactiveui rebuilt `ghcr.io/reactiveui/certum-signer:latest` on 2026-09-15 00:39 UTC, moving it from Ubuntu 24.04 to 26.04, jsign 7.4 to 7.5, and .NET 11 RC. Both 1.9.14 runs pulled the new image (`sha256:54d12a5f`):

| run | result |
|---|---|
| [34970816931](https://github.com/Rhizomatica/mercury/actions/runs/34970816931) | `missing required tool(s): python3`. Already fixed on trunk by 7ae152c. |
| [34973359409](https://github.com/Rhizomatica/mercury/actions/runs/34973359409) | SimplySign logs in and the session comes up, then `no signing key visible via SunPKCS11`, 3 of 3 attempts. |

`code-signing/` is unchanged since v1.9.13. The 2026-08-24 release signed with `sha256:fb03892c` (built 2026-07-01, Ubuntu 24.04, jsign 7.4).

## Change

Pin the sign job's container to `sha256:fb03892c`. I checked that the registry still serves that manifest. The python3 install from 7ae152c stays; on this image it is a no-op.

## To finish 1.9.14

Merge, then re-run the release. The version bump is idempotent.

## Not verified

This can't be tested before a real run, because the sign job needs the Certum secrets.

Moving to the 26.04 image is separate work, not done here. Upstream reactiveui/actions-common changed its own signing on 09-06 to retry `jsign` while the keystore has no certificate yet (27e4449). They report that the token lists before its objects are populated. Our `ss_key_alias` asks `keytool` once per attempt, with stderr discarded, so the real cause on 26.04 is unconfirmed.


rafael2k:
Superseded by c1718be: signing now works on the rebuilt certum-signer image without a pin, and 1.9.14 is released with signed binaries.

PR #300, opened 2026-09-20
Replace mongoose with our own WebSocket server (GPLv3 licence fix)

rafael2k:
Mongoose 7.20 is `GPL-2.0-only or commercial`. Mercury is GPL-3.0-or-later, and GPL-2.0-only cannot be combined with it — Cesanta dropped the "or later" deliberately, so there is no upgrade path. `mongoose.o` was linked into both the `mercury` binary and `libmercury_core.a`, which made every binary we ship undistributable and blocked Debian packaging. It was the only GPL-2.0-only file in the tree.

So Mercury now speaks HTTP and RFC 6455 itself: **28,820 lines of vendored library replaced by about 1,500 lines of ours.**

The public API (`ws_init`, `ws_broadcast_json`, `ws_broadcast_binary`, `ws_shutdown`) is unchanged, so `ui_communication.c` is untouched and the existing test stubs still match.

## New units

| File | Contents |
| --- | --- |
| `ws_server.c` | listener, poll loop, per-connection state, HTTP subset |
| `ws_frame.c` | RFC 6455 frame codec, pure functions |
| `ws_crypto.c` | SHA-1 and base64 for the handshake, from RFC 3174 / 4648 |
| `ws_json.c` | the UI command parser, lifted out of `mercury_websocket.c` |
| `ws_tls.c` | optional OpenSSL backend behind a seam |

## Cross-platform

Portability needed no new code. `common/os_interop.h` already supplies `poll()` on Windows (over `select()`, because WSAPoll is unreliable), `SOCK_CLOSE`, `sock_errno()` and the EAGAIN spellings; pthreads are used on all three platforms and `-lws2_32` was already in the Windows link line. There is no Windows or macOS `#ifdef` in the new server beyond `SO_NOSIGPIPE`.

## TLS

`wss://` now goes through OpenSSL (Apache-2.0, GPLv3-compatible), detected by pkg-config and so in practice Linux. The Windows and macOS builds vendor their dependencies statically and do not carry a TLS stack for a feature that is off by default; they serve plain `ws://` and refuse `wss://` at startup with a message saying so, rather than failing per connection. `make WS_TLS=0` builds without it anywhere.

Because of that, the web UI no longer guesses `wss://` first and falls back after two failed attempts — it uses the scheme the page was served over, which is what `waterfall.html` and `test.html` already did.

## Backpressure

Explicit now that we own the buffer: the spectrum stream is 2 KB every 50 ms, so a client that stops reading backs up fast. Past 256 KB queued we drop binary frames (the next one is along in 50 ms); control frames are never silently dropped, so past 2 MB we close the connection instead of lying about the client's state.

## Fuzzing

Splitting `ws_json.c` out unblocks `tests/fuzz/fuzz_ws_json.c`, deferred since it was written because reaching two `static` helpers meant linking the whole amalgamation. That parser is reachable by anyone who can open the UI port, which is unauthenticated and bound to 0.0.0.0.

## Packaging

`debian/copyright` claimed GPL-3+ for everything and never mentioned mongoose; it now also records iniparser and Unity as Expat. `debian/control` gains `libssl-dev` and `pkgconf`.

## Verification

- **409 unit tests pass**, including 16 new ones pinning the handshake digest against the RFC 6455 worked example, the three payload-length encodings, masking, and the frames we must refuse.
- **Fuzz harness clean over 400k mutated inputs** under ASan and UBSan.
- **End to end against the real binary on Linux** and, under wine, **against the mingw `mercury.exe`**: handshake with verified `Sec-WebSocket-Accept`, the 20 fps binary spectrum stream, tens-of-KB text frames, fragmented client messages, ping/pong and the close handshake all behave.
- **Shutdown with a browser-like client holding a socket open takes 103 ms**, well inside `main()`'s `alarm(10)` watchdog.
- **Backpressure confirmed**: a client that never reads had 1652 frames dropped over 3 minutes, with the connection kept and the process bounded.
- `mercury`, `mercury.exe`, `libmercury_core.a` and `mercury-ui` all build; no `mg_` symbol remains in any of them.

## One thing to apply separately

CI installs no `libssl-dev`, so released Linux packages would quietly lose `wss://`. That change touches `.github/workflows/`, which my token cannot push — it is stashed locally and needs applying by hand.

## Unrelated finding

While measuring backpressure I found Mercury's RSS grows ~31 kB/s (~2.7 GB/day) **with the UI switched off entirely**, so it predates this work and is not in this code path. Worth a separate look, since field stations run for days.


pedromessetti (commented):
Reviewed. Strong change overall — the licensing rationale is correct and material, the module split (codec / crypto / JSON / TLS / server) is clean, and the test coverage is real. I also verified the build-matrix logic: `make windows` sets `OS=Windows_NT` so the config.mk guard genuinely skips OpenSSL detection for the mingw build, and the arm64/armhf jobs set `PKG_CONFIG_LIBDIR`/`PKG_CONFIG_ALLOW_CROSS` so they resolve the cross-arch libssl.

Findings:

1. **Handshake doesn't validate `Connection: Upgrade`** — `gui_interface/websocket/ws_server.c:631`. Only `Upgrade`, `Sec-WebSocket-Version`, and `Sec-WebSocket-Key` are checked; RFC 6455 §4.2.1 requires confirming the `Connection` header carries the `Upgrade` token. Low impact (browsers always send it), but it's a spec gap and `ws_conformance.py` doesn't cover it.

2. **`ws_json_sanitise_utf8` isn't a full UTF-8 validator** — `gui_interface/websocket/ws_json.c`. It only checks continuation-byte shape, so overlong encodings, UTF-16 surrogates (`0xED 0xA0..0xBF`), and code points above U+10FFFF (`0xF4 0x90+`) pass through; a strict browser will still reject such frames. The header comment overstates what it does.

3. **`ws_json_find_key` matches `"command"` inside string values** — `gui_interface/websocket/ws_json.c`. The scanner is quote/escape-unaware, so e.g. `{"value":"\"command\":\"evil\"","command":"real"}` resolves the wrong key. This is the exact entry point being fuzzed (unauthenticated port), so worth tightening.

Minor / non-blocking:
- `ws_server_client_count` (`ws_server.c`) and `ws_asset_name` (`pack_web.sh`) are dead code; the former also reads `s->conns` lock-free (fine only because it's unused).
- `conn_close_ws` echoes the client's close code verbatim, including reserved-on-the-wire codes (1005/1006/1015).
- The PR description still says the CI `libssl-dev` change "needs applying by hand", but the diff already includes it (`.github/workflows/*`).

rafael2k:
Thanks Pedro — all three findings and both minors were real, and they're fixed in cd593ff. One of the three was right about the cause but wrong about the effect, so I want to be precise about it.

**1. `Connection: Upgrade` — fixed.** You were right, and slightly worse than you put it: `header_has_token()` already existed for exactly this purpose and I'd only ever applied it to `Upgrade`. Now validated, with two conformance checks (rejected without the token, accepted in a `keep-alive, Upgrade` list).

**2. `ws_json_sanitise_utf8` — fixed.** Confirmed before changing anything: all four classes went through untouched.

```
overlong '/' (C0 AF)          C0 AF          <-- unchanged
overlong NUL (E0 80 80)       E0 80 80       <-- unchanged
UTF-16 surrogate (ED A0 80)   ED A0 80       <-- unchanged
above U+10FFFF (F4 90 80 80)  F4 90 80 80    <-- unchanged
```

Now range-checked per lead byte, with the boundaries (U+D7FF, U+10FFFF) still passing. This one has a practical bite I hadn't appreciated: the status frame is built from soundcard and rig names we don't control, so a single odd device name would have made a strict client fail the connection repeatedly.

**3. `ws_json_find_key` — fixed, but your example doesn't reproduce.** `{"value":"\"command\":\"evil\"","command":"real"}` parses as `real`. Inside a JSON string the quotes are escaped, so the bytes are `\"command\"` — and the needle `"command"` includes the closing quote, which meets a backslash instead. No match, so it falls through to the real key.

Your diagnosis of the *cause* was right though — the scanner was quote- and structure-unaware — and once I went looking with that in hand, here's what it actually did:

```
{"command":"x","value":"a \"b\" c"}                    -> value = 'a \'
{"command":"set_ptt_config","value":"/dev/tty\"USB0"}  -> value = '/dev/tty\'
{"value":{"command":"evil"},"command":"real"}          -> command = 'evil'
```

So the live defect was **truncation**, not key confusion: any device path or callsign containing a quote got cut at the backslash. Plus the nested-key shadowing you predicted, which is latent for our flat protocol but real. It's now a small proper parser — skips nested objects and arrays, honours escapes, and decodes what a browser actually emits (`\" \\ \/ \b \f \n \r \t` and `\uXXXX` including surrogate pairs).

**Minors** — all confirmed, all done. `ws_server_client_count` and `ws_asset_name` had zero callers and are gone (the former did read `s->conns` unlocked, as you noticed). Close codes 1005/1006/1015 are no longer echoed; RFC 6455 §7.4.1 is explicit that they must not appear in a frame. And the stale "needs applying by hand" line in the description is corrected — that change is in the diff.

**Verification:** 419 unit tests (7 new, pinning escape decoding, the nested-key case, whole-key matching, malformed input, and both halves of the UTF-8 range check). Conformance 26/26, up from 20, against an ASan+UBSan build with no sanitizer report. The rewritten parser is clean over 400k mutated inputs under ASan/UBSan, with two new corpus seeds for the escape and `\u` paths.

Also noted your independent check of the build matrix — that's exactly the part I was least able to verify locally, so thank you for going through it.

